PR #55, opened 2020-09-17
pluma: clarify YAML files errors

Pixep:
Includes the original error message in the final message


mjftw (approved):
Looks pretty harmless

PR #56, opened 2020-09-18
replaced login_pass with system.credentials.password

Zematus:
this is the change suggested by @Pixep to solve the problem with ssh not receiving the proper credentials from the CLI


Pixep (approved):
Good to go!

PR #58, opened 2020-09-19
Feature/shell test wait for completion

Pixep:
* For shell commands, wait for command completion, instead of console silence
* Reorganize methods to have less responsibility (just a start)
* Move CommandRunner to a dedicated file


pluma/test/shelltest.py
```diff
@@ -52,41 +55,87 @@ def test_body(self):
...
+        if not matched:
+            CommandRunner.log_error(test_name=test_name, sent=command, output=output,
+                                    error='No response within timeout, or failed device'
+                                    ' failed to send return code. If this is not running'
+                                    ' in a shell, set "runs_in_shell" to "false".')
+
+        output = re.sub(re.escape(matched) + r'$', '', output)
+        retcode_match = re.search(retcode_token+r'(-?\d+)', matched)
+        if not retcode_match:
+            raise Exception('Failed to find return code value')
```
mjftw: Would maybe be better to raise a named exception.
E.g.
https://github.com/Witekio/pluma-automation/blob/22c60cf27ee75114523d66f973123b3e42f20fa6/pluma/test/test.py#L51

mjftw (approved):
Looks great 👍🏼 
One minor comment, but it's not worth blocking the PR for.

pluma/test/commandrunner.py
```diff
@@ -0,0 +1,173 @@
...
+                CommandRunner.log_error(test_name=test_name, sent=command, output=output,
+                                        should_print=should_print,
+                                        error='Response did not match expected')
+            else:
+                message = CommandRunner.format_command_log(sent=command, output=output,
+                                                           should_print=should_print)
+                log.log(message)
+
+    @staticmethod
+    def cleanup_command_output(command: str, output: str) -> str:
```
mjftw, line 126: Found a bug here when trying to run a new shell test that runs `[ -e "/bin/bash" ]` as the script with `run_on_host=True`.
Inputs to function (found in debugger): `'command: [ -e "/bin/bash" ] ; echo pluma-retcode=$?, output: '[ -e "/bin/bash" ] ; echo pluma-retcode=$?\r\n''`

The function strips out all of the output and returns and empty string.
This then causes an error to be raised at:
```python

        if not output:
            CommandRunner.log_error(test_name=test_name, sent=command, output=output,
                                    error='No response received after sending command')
```

mjftw (changes requested):

pluma/test/commandrunner.py
mjftw: [See screenshot](https://pasteboard.co/JspG9Sw.png)

mjftw: Found on WIP branch: https://github.com/Witekio/pluma-automation/tree/feature/testsuite-file-tests

mjftw: This fixed it for me. I think that the assumption that all commands must produce output text, and not just a return code is wrong.

```diff
diff --git a/pluma/test/commandrunner.py b/pluma/test/commandrunner.py
index 06acac0..9c26027 100644
--- a/pluma/test/commandrunner.py
+++ b/pluma/test/commandrunner.py
@@ -12,7 +12,7 @@ class CommandRunner():
     def run(test_name: str, console: ConsoleBase, command: str, timeout: int = None):
         '''Run a command in a Shell context'''
         retcode_token = 'pluma-retcode='
-        command += " && echo pluma-retcode=$?"
+        command += " ; echo pluma-retcode=$?"
         output, matched = console.send_and_expect(
             command, timeout=10, match=retcode_token+r'-?\d+')
 
@@ -41,12 +41,8 @@ class CommandRunner():
 
         output = CommandRunner.cleanup_command_output(command, output)
 
-        if not output:
-            CommandRunner.log_error(test_name=test_name, sent=command, output=output,
-                                    error='No response received after sending command')
-        else:
-            message = CommandRunner.format_command_log(sent=command, output=output)
-            log.log(message)
+        message = CommandRunner.format_command_log(sent=command, output=output)
+        log.log(message)
 
         return output
```

Pixep: Yeah that's right, modified the code accordingly

pluma/test/shelltest.py
Pixep: Done

mjftw (approved):

PR #64, opened 2020-09-20
Feature/uhubctl power control

Pixep:
Add power control via uhubctl utility for USB hubs


pluma/core/power/uhubctl.py
```diff
@@ -0,0 +1,50 @@
...
+class Uhubctl(PowerBase):
+    ''' USB hub power control using uhubctl utility '''
+
+    def __init__(self, location: str, port: int, reboot_delay: int = None):
+        PowerBase.__init__(self, reboot_delay)
+        self.location = location
+        self.port = port
+
+        if not location or not port:
+            raise Exception('Uhubctl: "location" and/or "port" '
```
mjftw, line 17: Would be better to raise a PDUError here.
https://github.com/Witekio/pluma-automation/blob/3265febdf77a33cc74f92eb888138040cc24b4dc/pluma/core/pdu.py#L11

Now that I look at it, maybe these should be called `PowerError` instead, nobody knows that `PDU` stands for Power Distribution Unit. Something to be put on another issue maybe.

mjftw (approved):
I've never used uhubctl before, looks pretty cool!

Since we're adding a new power controller, it would be good to add some reference to this in the docs (maybe in supported hardware).

mjftw (changes requested):
Sorry meant to stick it as request changes for the Exception thing

pluma/core/power/uhubctl.py
mjftw: https://github.com/Witekio/pluma-automation/blob/master/docs/style-guide.md#exceptions

Pixep: Good point. Changed to `PDUError`

Pixep:
> Since we're adding a new power controller, it would be good to add some reference to this in the docs (maybe in supported hardware).

Added to the list!

mjftw (approved):

PR #65, opened 2020-09-20
pluma: switched multipled verbose logs to debug

Pixep:
Verbose is used to gain information regarding the operations at a high
level, and these were mostly unhelpful for a regular user.


mjftw (approved):
Looks good :)

PR #68, opened 2020-09-20
Add variables definition and subtitution support

Pixep:
Variables can be defined in the target file, and will be substitued in
the tests file only.


tests/generic/cli/test_ConfigPreprocessor.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+from pluma.cli import PlumaConfigPreprocessor
+
+
+def test_PlumaConfigPreprocessor_should_keep_original_with_no_variables():
+    preprocessor = PlumaConfigPreprocessor(None)
+    origina_content = 'content: value\notherline\nan a new one.'
```
mjftw, line 8: Typo: ~~origina_content~~ original_content

pluma/cli/configpreprocessor.py
```diff
@@ -0,0 +1,33 @@
...
+        self.variables = variables
+
+    def preprocess(self, raw_config: str) -> str:
+        def token_to_variable(token):
+            # remove '${' and '}'
+            return token[2:len(token)-1]
+
+        missing_variables = []
+        vars_found = re.findall(r'\${\w+}', raw_config, flags=re.MULTILINE)
+        vars_found = map(token_to_variable, vars_found)
```
mjftw, line 17: I'd probably just put this one one line rather than redefining vars_found:
```python
vars_set = set(map(token_to_variable, vars_found))
```

And maybe rename it to something like `unique_vars` as I'm guessing that's what the set is being used for.

pluma/cli/configpreprocessor.py
```diff
@@ -0,0 +1,33 @@
...
+from .config import ConfigPreprocessor
+
+
+class PlumaConfigPreprocessor(ConfigPreprocessor):
+    def __init__(self, variables: dict):
+        self.variables = variables
+
+    def preprocess(self, raw_config: str) -> str:
+        def token_to_variable(token):
+            # remove '${' and '}'
```
mjftw, line 12: Doesn't really matter as this is a nested function, but would be more typical to put this as a `'''` docstring rather than a `#` comment.

pluma/cli/config.py
```diff
@@ -119,21 +121,29 @@ def selected_tests(self, config: Configuration) -> list:
         return list(filter(lambda test: (test.selected), self.all_tests(config)))
 
 
+class ConfigPreprocessor(ABC):
```
mjftw, line 124: Using `ABC` here over `metaclass=ABCMeta` is inconsistent with the rest of the code.

But actually I think it's more readable and therefore better since it does the same.
The right thing to do is probably to swap the rest of the code over.

mjftw (changes requested):
Looks good.
Only minor comments. Quick fix then this is good to go.

tests/generic/cli/test_ConfigPreprocessor.py
Pixep: Done

pluma/cli/configpreprocessor.py
Pixep: Yep, makes sense. Done.

pluma/cli/configpreprocessor.py
Pixep: Sure why not :)

pluma/cli/config.py
Pixep: Ok, this is consistent but only in `cli` yeah. Will create an issue, as both seem slightly different, and I prefer no to cause regressions here.

Pixep: https://github.com/Witekio/pluma-automation/issues/70

mjftw (approved):

mjftw:
All good from my point of view!
Feel free to merge once the conflict is resolved.

Issue #41, opened 2020-09-10
Test suite: Test that a file exists


Pixep:
file check: exists, type, size, ...

PR #74, opened 2020-09-23
Feature/testsuite file tests

mjftw:
Closes: https://github.com/Witekio/pluma-automation/issues/41

Note, currently these cause the following type errors:
```
(venv) merlin@pangolin:~/projects/pluma-automation$  cd /home/merlin/projects/pluma-automation ; /usr/bin/env /home/merlin/projects/pluma-automation/venv/bin/python3.6 /home/merlin/.vscode/extensions/ms-python.python-2020.8.109390/pythonFiles/lib/python/debugpy/launcher 37559 -- -m pluma -c /home/merlin/projects/pluma-automation/pluma.yml -t /home/merlin/projects/pluma-automation/pluma-target.yml 
No console defined in the device configuration file
The following configuration attributes were not recognized or not used:
{'testsuite.filesystem.CheckFileSize': {'path': '/bin/bash', 'min': 10, 'max': 10000, 'run_on_host': True}}
```

This is thought to be due to an existing bug, but must still be fixed before merging.
This PR also relies on bug fixes mentioned in comments on https://github.com/Witekio/pluma-automation/pull/58.

This PR is a work in progress, as it would be good to be able to specify file size like "10kb", "100mb", "2gb" etc, rather than having to use the size in bytes.

It would also be good to have more informative output when a test fails.


Pixep:
Oh that's great. I think it works out even without the kb/mb and all. Is the error you're seeing only with venv?

mjftw:
> Oh that's great. I think it works out even without the kb/mb and all. Is the error you're seeing only with venv?

Oh okay great! So it's working for you?
I think the issues I was seeing are not to do with these tests, so maybe something to do with virtualenv but not sure how.

Pixep:
> Oh okay great! So it's working for you?

Haha no, not tested :smile: , but your problem looks like that other PR you opened. Basically since your `--plugin` PR, we always need to add them to an `__init.py__`, which is... nice, but definitely something to not miss when creating tests, can be frustrating.

mjftw:
I'm holding off moving this out of draft until I have unit tests for all these tests.
Should be pretty easy to do using run_on_host in the tests, but need to get all the testing framework helpers in for this first.

Pixep:
Good to go once rebased I believe @mjftw ! It shows commits previously on master in the commit diff, so waiting to merge to avoid duplicates or history weirdness.

mjftw:
> Good to go once rebased I believe @mjftw ! It shows commits previously on master in the commit diff, so waiting to merge to avoid duplicates or history weirdness.

Oops, looks like I'd forgotten to change the merge base from master to the branch that had some fixes that this PR was based off.
The other PR got merged, but the base for this PR didn't changed to master because it was already set to that, leaving duplicate commits.
Rebased on master now and they've gone away.

Nice spot!

Pixep (approved):
All good

PR #77, opened 2020-09-23
added copy to device action

Zematus:
I couldn't fully test it because networking is currently broken on the device with the latest image. Anyone wan't go give it a shot?

Here's an example action:

- copytodevice:
    file: 'test.txt'
    destination: '/tmp/'
    timeout: 10


mjftw (approved):
Nice one!

Tested and working on my machine. Was able to copy a file from one directory on host to another by setting up the SSH console to point back to the same machine.
Definitely not the simplest way to copy a file, but proves it works :)

Issue #75, opened 2020-09-23
Pluma --plugins flag behaves unexpectedly


mjftw:
**Describe the bug**
The Pluma `--plugin` flag is not working as intended, and gives some rubbish error messages.

The plugins directory specified should have the following structure:
`<plugindir>/module_name/<filename>.py`
and `<filename>.py` should contain the tests.

The error given when the plugins directory has the following structure is awful:
`<plugindir>/<filename>.py`
and `<filename>.py` should contain the tests.

(Probably this structure should be allowed anyway)
```
(venv) merlin@pangolin:~/projects/pluma-automation$ pluma -c ./pluma.yml -t ./pluma-target.yml --plugin ./plugindir/ --debug check
No console defined in the device configuration file
Components:
    Serial:  None
    SSH:  None
    Prompt:  \S+@\S+:
    Login:  mendel
    Password:  ******
    Power control:  None
    Storage:  None
    USB Hub:  None

Traceback (most recent call last):
  File "/home/merlin/projects/pluma-automation/pluma/__main__.py", line 87, in main
    check_only=True)
  File "/home/merlin/projects/pluma-automation/pluma/cli/api.py", line 26, in execute_run
    target_config_path, show_tests_list=check_only)
  File "/home/merlin/projects/pluma-automation/pluma/cli/api.py", line 79, in build_test_controller
    testsConfig = TestsConfig(tests_config, Pluma.tests_providers())
  File "/home/merlin/projects/pluma-automation/pluma/cli/testsconfig.py", line 34, in __init__
    self.__populate_tests(config)
  File "/home/merlin/projects/pluma-automation/pluma/cli/testsconfig.py", line 85, in __populate_tests
    self.tests = self.tests_from_sequence(sequence, self.test_providers)
  File "/home/merlin/projects/pluma-automation/pluma/cli/testsconfig.py", line 140, in tests_from_sequence
    supported_actions=supported_actions)
  File "/home/merlin/projects/pluma-automation/pluma/cli/testsconfig.py", line 156, in tests_from_action
    tests = provider.all_tests(key=action_key, config=Configuration(action_config))
  File "/home/merlin/projects/pluma-automation/pluma/cli/pythontestsprovider.py", line 31, in all_tests
    all_tests = self.find_python_tests()
  File "/home/merlin/projects/pluma-automation/pluma/cli/pythontestsprovider.py", line 64, in find_python_tests
[...]

PR #79, opened 2020-09-24
Add CLI --plugin tests in cli unit test folder

mjftw:
This PR adds tests which show the failure behaviour mentioned in https://github.com/Witekio/pluma-automation/issues/75.

It is a WIP and further commits will fix the issue, rather than just highlight it.


mjftw:
@Pixep , can you see anything wrong with the config files or test I've created that would cause the following error?
```
Pluma CLI exited with code -1 
stdout: No console defined in the device configuration file
Some include/exclude rules matched no tests: ['example_plugin.mytests']
```

Pixep:
Thanks for the unit test @mjftw. Can you move it under `tests/generic/cli`? There's now `pluma.py` there, and this would probably make more sense to be in this file, or at least under `cli`.

> @Pixep , can you see anything wrong with the config files or test I've created that would cause the following error?

Sure I can have a look

mjftw:
> Thanks for the unit test @mjftw. Can you move it under `tests/generic/cli`? There's now `pluma.py` there, and this would probably make more sense to be in this file, or at least under `cli`.
> 
> > @Pixep , can you see anything wrong with the config files or test I've created that would cause the following error?
> 
> Sure I can have a look

Sure, good thinking.
Done.

Pixep:
Fixed, I pushed a quick commit. Feel free to squash. This is basically normal, no test was found, because the __init__.py was not including your class

mjftw:
> Fixed, I pushed a quick commit. Feel free to squash. This is basically normal, no test was found, because the `__init__.py` was not including your class

Ah nice one 👍🏼

mjftw:
> Fixed, I pushed a quick commit. Feel free to squash. This is basically normal, no test was found, because the `__init__.py` was not including your class

Ah okay makes sense.
I think if we could find a way where this doesn't need to be done for every new test added then that would be good.
I can see this tripping people up a lot in future when adding new tests, and it means the `__init__.py` will have to be modified a lot,  potentially causing a lot of merge conflicts as people add more tests.

PR #80, opened 2020-09-24
ShellTest: 'login_automatically' now defaults to False

Pixep:
Causes confusing error in various cases, better to keep it separated by
default.


mjftw (approved):
Sounds like a good idea. Doing less by default is usually less confusing.
One conflict to fix since I merged a different PR, then feel free to merge

PR #82, opened 2020-09-24
pluma: substitue variables from environment

### LINKED ISSUE #17: Support environment variables replacement in configuration files
Support using `${VARIABLE_NAME}` tokens in the target and test configuration files, replaced with environment variables content.

Pixep:
Substitute variables defined as ${} in both configuration files. Closes 
Fixes #17


README.md
```diff
@@ -257,6 +257,14 @@ Supported attributes:
   * `- powercycle:` Use the power controller defined to power cycle the board (off and on)
     * `off_duration_ms:` Duration for which the device is powered off. Actual duration may be longer
 
+### Variables and substitution
+
+Variables are used in both the target and tests configuration files as `${my_var}`. Variables can be defined either:
+* In the target configuration
+  * These are defined, under `variables` (see target definition), and substituted only in the tests configuration
+* As environment environment variables
+  * Environment variables always take precedence over target definition variables, and can be used in both the target, and test configuration files. These are typically set with `EXPORT my_var=value` on Linux platforms.
```
mjftw, line 266: Feels to me like it make more sense for variables defined in target config to overwrite env, since the target conf ones are specific to that board, whereas the env vars are global and less specific.
That would be my expected behaviour as a user.
I think the env is more likely to have leftover variables that people have set and forgotten about that cause odd errors by overwriting the local ones the user has just set.

mjftw: Might be good to put something in the verbose/debug output mentioning the substitutions made and where they sourced from. Or maybe just a warning message when the same env var is detected locally in target config and globally in env.
I can see this being a cause of frustration for users when their env isn't used without them knowing and the tests behave wrong.

tests/generic/cli/test_Pluma.py
```diff
@@ -1,16 +1,23 @@
 from os import path
+import os
 import pytest
 
 from pluma.cli import Pluma, ConfigurationError, TestsConfigError, TargetConfigError
 
 
+config_folder = path.join(path.dirname(__file__), 'test-configs/')
```
mjftw, line 8: Might be good for us to put this somewhere accessible by multiple test files as I can see this being needed in a few of them. Doing so would enforce consistency throughout the test files.
Maybe passed in as a fixture? `configs_dir`?
Also, this doesn't matter so much but I prefer `dir` or `directory` over `folder` as it's more commonly used in Linux and it's primarily a Linux tool. But just personal preference.

tests/generic/cli/test_Pluma.py
```diff
@@ -1,16 +1,23 @@
...
+
+
+def config_file_path(config: str):
```
mjftw, line 11: Could be a fixture that returns this function, with the config folder wrapped into it in `conftest.py`

tests/generic/cli/test_Pluma.py
```diff
@@ -1,16 +1,23 @@
...
+    return path.join(config_folder, f'{config}.yml')
+
+
 def run_all(test_file: str, target_file: str):
```
mjftw, line 15: Nothing to stop this being a fixture which takes the previous fixture as a dependency. Not sure about this though. Thoughts?

mjftw (changes requested):
The code all looks good to me, only comments are more about behaviour.

README.md
Pixep: So in the solutions I know (like AWS) env vars do override file config, because environment variables are more versatile than files. Hence you have Env vars > Local config > global config > default parameters. In GitLab for instance, you can use the same repo and files, but run the same builds with different environment variables. So you need environment variables to have a higher priority. Not sure about Yocto? Do have other examples? And yeah can add something to the --verbose mode to print vars.

tests/generic/cli/test_Pluma.py
Pixep: Yeah it does work. Added here: https://github.com/Witekio/pluma-automation/issues/95

README.md
mjftw: Oh yeah fair enough, if that's the done thing then let's do that.
A warning about redefined variables would still be nice though.

tests/generic/cli/test_Pluma.py
Pixep: Added here: #95

tests/generic/cli/test_Pluma.py
Pixep: Added here: #95

mjftw (approved):

PR #84, opened 2020-09-25
CommandRunner: show only base command when command succeeds

Pixep:


mjftw (approved):
Looks good, much better user experience.

PR #85, opened 2020-09-25
pluma.test: split test.py in different files

Pixep:
`test.py` was way to big, step toward simplifying the whole architecture


mjftw:
> `test.py` was way to big, step toward simplifying the whole architecture

Completely agree, all of the test framework files and classes are way too heavy.
I've wanted to split them apart for ages so thanks for this!

pluma/test/__init__.py
```diff
@@ -1,8 +1,11 @@
-from .exceptions import *
-from .test import TestBase, TestCore, TestRunner
-from .unittest import deferred_function
-from .testcontroller import TestController
-from .commandrunner import CommandRunner
-from .shelltest import ShellTest
-from .executabletest import ExecutableTest
+from .exceptions import TestingException, TaskFailed, AbortTesting, \
+    AbortTestingAndReport  # noqa: F401
```
mjftw, line 2: What are all these `# noqa: F401` lines about?

pluma/test/testrunner.py
```diff
@@ -1,233 +1,19 @@
...
-        self.board.log("\n=== POST BOARD OFF ===", color='blue', bold=True)
-
-    def report(self):
-        self.board.log("\n=== REPORT ===", color='blue', bold=True)
+from pluma.test import TestCore, TestBase, TestingException, \
+    AbortTesting, AbortTestingAndReport
 
 
 class TestRunner():
+    """Orchestrates tests globally"""
```
mjftw, line 15: Might be worth mentioning that It only orchestrates a set of tests to run a single time.
Only half the picture with `TestController`.

mjftw (changes requested):

pluma/test/__init__.py
Pixep: Yep, ignored it in Codacy

pluma/test/testrunner.py
Pixep: Updated

mjftw (approved):
Looks great 👍🏼

PR #86, opened 2020-09-25
TestController: start splitting into standalone classes

Pixep:
@mjftw : Is this something you can test?


mjftw (approved):
So much better!
I've wanted to do this for a long time now, thanks!

PR #92, opened 2020-09-25
Closes #75, closes #90 Feature/testuite doesnt need all tests in init

### LINKED ISSUE #75: Pluma --plugins flag behaves unexpectedly

### LINKED ISSUE #90: Make it so `__init__.py` doesn't need to include every test class
> Fixed, I pushed a quick commit. Feel free to squash. This is basically normal, no test was found, because the `__init__.py` was not including your class

Ah okay makes sense.
I think if we could find a way where this doesn't need to be done for every new test added then that would be good.
I can see this tripping people up a lot in future when adding new tests, and it means the `__init__.py` will have to be modified a lot,  potentially causing a lot of merge conflicts as people add more tests.

_Originally posted by @mjftw in https://github.com/Witekio/pluma-automation/pull/79#issuecomment-698850474_

mjftw:
With this we can now find tests in scripts either in the root `--plugin` dir (and in `pluma.plugins`) or in submodules.
For tests in submodules, you now only need to import the submodule in the module `__init__` init, not every test within that submodule. Tests in submodules are found automatically.

This PR also fixes an existing bug I found where `TestBase` based classes imported into the `mytests.py` test module were also found and assumed to be real runnable tests.
E.g.
mytests.py
```python
from pluma.test import ShellTest
```

Previously `ShellTest` this would have been reported as being one of the tests in the `mytests` module, so could have been called in the yaml with
```yaml
sequence:
- core_tests:
    include: [mytests]
    parameters:
      mytests.ShellTest:
        name: Merlin
      topleveltests.World:
```

This bug is also what was causing the error `ValueError: substring not found` seen in https://github.com/Witekio/pluma-automation/issues/75. This PR effectively closes that issue as well.


Pixep:
Quick question: In our networking tests, we have `NetworkingTestBase`. Is this one exposed automatically now? Because it shouldn't, as this is a base test only. Any idea to exclude it, if so?

mjftw:
> Quick question: In our networking tests, we have `NetworkingTestBase`. Is this one exposed automatically now? Because it shouldn't, as this is a base test only. Any idea to exclude it, if so?

I'd say yes it'll be exposed, hadn't considered base classes. Hmmmm
1. Well one way around this would be to rely on naming convention. E.g. if the class name ends with "Base" it's not imported.
I'm a massive fan of relying on naming convention though.
2. A different way would be to check if the class is subclassed, and not include it if it is, but this has the disadvantage of silently breaking inclusion for base classes that people intend to make instances of.
3. We could add an optional `__all__` array (or some other name) to the file denoting all of the classes that should be found, but adds complexity. 
4. We could leave some marker on the individual base classes in whatever form (be it attributes, functions, inheritance etc etc), but that then requires the test class be modified, which I don't like.
5. We could go a similar route that we have with DeviceActions with registering them and remove the custom code to check for TestBase type classes. If we were to abstract the registration code then this could be a nice scalable solution for other user expanded things going forward.

So options I can think of are rely in naming convention or only include classes that are leaf classes.

mjftw:
> > Quick question: In our networking tests, we have `NetworkingTestBase`. Is this one exposed automatically now? Because it shouldn't, as this is a base test only. Any idea to exclude it, if so?
> 
> I'd say yes it'll be exposed, hadn't considered base classes. Hmmmm
> 
>     1. Well one way around this would be to rely on naming convention. E.g. if the class name ends with "Base" it's not imported.
>        I'm a massive fan of relying on naming convention though.
> 
>     2. A different way would be to check if the class is subclassed, and not include it if it is, but this has the disadvantage of silently breaking inclusion for base classes that people intend to make instances of.
> 
>     3. We could add an optional `__all__` array (or some other name) to the file denoting all of the classes that should be found, but adds complexity.
> 
>     4. We could leave some marker on the individual base classes in whatever form (be it attributes, functions, inheritance etc etc), but that then requires the test class be modified, which I don't like.
> 
>     5. We could go a similar route that we have with DeviceActions with registering them and remove the custom code to check for TestBase type classes. If we were to abstract the registration code then this could be a nice scalable solution for other user expanded things going forward.
> 
> 
> Thoughts?

Either way, this fixes an existing bug, so if we don't want to merge this right away we should probably still keep the fix. Maybe via a new PR.

Pixep:
> We could go a similar route that we have with DeviceActions with registering them and remove the custom code to check for TestBase type classes. If we were to abstract the registration code then this could be a nice scalable solution for other user expanded things going forward.

Yeah, that's probably what I would lean towards. We can already integrate that, once rebased. Currently it includes commits already on master (but yet different), so you'll have to `rebase --onto` :)

mjftw:
> > We could go a similar route that we have with DeviceActions with registering them and remove the custom code to check for TestBase type classes. If we were to abstract the registration code then this could be a nice scalable solution for other user expanded things going forward.
> 
> Yeah, that's probably what I would lean towards. We can already integrate that, once rebased. Currently it includes commits already on master (but yet different), so you'll have to `rebase --onto` :)

Okay, I'll go with that.

One thing to note is that even without these changes, a user could still add a python file in the root of their `--plugin` dir, and if it contained any `TestBase` based classes they would be imported.

I think for the `--plugin` dir, being able to put the test scripts in the root of the dir makes a lot more sense than having to nest them inside another dir. Does mean that the issue you mentioned is still here though.

mjftw:
I've done a pretty manual rebase onto master with this branch (using cherry-picks and and hard resets), edited the branch history to split up a commit into stuff that adds bugfix and stuff that adds new features, and finally I've created a new branch and PR from just the bugfix bits.

Pixep (approved):
Thanks, good one

PR #94, opened 2020-09-25
cli: always run in sequential mode

Pixep:
Parallel mode is not supported by the CLI


mjftw (approved):
Yeah makes sense to force that.
Longer term we might want to think about whether parallel mode is needed at all, and if it is, how it can be simplified.

PR #99, opened 2020-09-27
Fix various linter warnings throughout the codebase

Pixep:


pluma/core/hub.py
```diff
@@ -63,11 +63,16 @@ def _pack_devinfo(self, device):
...
         return devinfo
 
-    def filter_downstream(self, filters={}, excludes={}):
+    def filter_downstream(self, filters: list = None, excludes: list = None) -> list:
         return self._filter_dictarr(
             filters=filters, excludes=excludes, dictarr=self.downstream)
 
-    def _filter_dictarr(self, filters={}, excludes={}, dictarr=[{}]):
+    def _filter_dictarr(self, filters: list = None, excludes: list = None,
+                        dictarr: list = None) -> list:
```
mjftw, line 71: This could be List[Dict]] if we wanted to get really fancy, but not important.

mjftw (dismissed):
Lots of great small improvements here :)

pluma/core/hub.py
Pixep: Yeah I think we will need to do that everywhere, once and for all. I tried to keep the changes small to avoid breaking something

Pixep: Added here: https://github.com/Witekio/pluma-automation/issues/107

Pixep:
Merge after a rebase only

PR #101, opened 2020-09-27
README.md: separate CLI and packages documentation

Pixep:
The README is a bit confusing right now, as it mixes concepts from both the CLI and pluma packages. Some effort to reorganize it a bit better.


mjftw (dismissed):
Great idea, and the new code example is much less rubbish.

README.md
```diff
@@ -3,47 +3,15 @@
 ![Native installation and tests](https://github.com/Witekio/pluma-automation/workflows/Native%20installation%20and%20tests/badge.svg)
 ![Docker Image CI](https://github.com/Witekio/pluma-automation/workflows/Docker%20Image%20CI/badge.svg)
 
-Pluma Automation (formerly Automation Lab) is a tool created by [Witekio](https://www.witekio.com) to perform black box testing of embedded hardware; designed to be as lightweight as possible!
+Pluma Automation (formerly Automation Lab) is a tool created by [Witekio](https://www.witekio.com) to perform automated testing on embedded devices. It has no requirements on the device side, but was primarily design for Linux-based systems. This tools is can be used in two different ways:
```
mjftw, line 6: To make it really clear that it's very general purpose, you could rephrase this line a little.
"It was primarily designed for Linux-based devices, but there is no requirement for Linux and it should work with any device." Or something like that.
Basically to make it clear to people not running Linux on their target that the tool is still useful to them.

Pixep: Hmm... yeah that's true for device control, but not for the test suite and all. Or maybe we need to move our tests to a `testsuite.linux.memory` etc.

Pixep: Removed the "but was primarily design for Linux-based systems" altogether. This can be specified in specific testsuite tests directly.

mjftw (approved):

PR #105, opened 2020-09-29
Added pytest fixture to allow use of temporary files in tests

mjftw:
This should allow for the Pluma CLI to be tested using config file contents
declared inline as strings, rather than having to have an actual file exist
for each test (scales badly).

The intention is for all new tests that need to use config files use this
fixture instead of real filesystem files.


Pixep (approved):
Looks good thanks. I think in some situations we definitely want to have the option to use a specific file, but I like what this will allow to do!

PR #115, opened 2020-10-01
tests: fix make test, runs with coverage

Pixep:
`make test` should not run coverage


mjftw (approved):
Huh, I didn't realise that it was running coverage before.

PR #117, opened 2020-10-02
Bugfix/core tests class finder is too greedy

mjftw:
Split from https://github.com/Witekio/pluma-automation/pull/92, following discussions in the comments: https://github.com/Witekio/pluma-automation/pull/92#issuecomment-702161373


pluma/cli/config.py
```diff
@@ -89,6 +89,9 @@ def __init__(self, name: str, testclass: type, test_provider: object,
             raise ValueError(
                 f'Parameter sets for test "{name}" should be a list of dictionaries')
 
+    def __repr__(self):
+        return f'[{self.__class__.__name__}] {self.name}: {self.parameter_sets}'
```
Pixep, line 93: Only comment here: The TestBase inherited tests I created so far appear as `module.class[parameters]`, so it would probably make sense to use the same naming convention. And if the test has no meaningful parameters, the `[parameters]` is not used in the naming. I think it follows-ish original Pluma tests naming as well/

Pixep (commented):
One small comment to discuss

Pixep (commented):
Minor comment/question regarding naming

pluma/cli/config.py
mjftw: Current: `[TestDefinition] setup: [{'script': 'echo "Setting up!"', 'run_on_host': True, 'name': 'setup'}]`
Requested changes: `pluma.cli.config.TestDefinition[{'script': 'echo "Setting up!"', 'run_on_host': True, 'name': 'setup'}]`

I was a little apprehensive as I think it's definitely good to have the name in there, but actually the parameter sets seems to have the name anyway.

I've made the requested changes.

Pixep: Oh I'm sorry, this is only the TestDefinition, and that's within the `[]` okay... Yeah that's fine as it. Your change didn't seem to get pushed, but I'm good with what you had then anyway!

Pixep (dismissed):

PR #118, opened 2020-10-02
Fix bug in pluma config fixture

mjftw:


Pixep (approved):
Agree with the changes, and all checks are good

PR #119, opened 2020-10-05
Feature/targetconfig warn on shadowed variables

### LINKED ISSUE #97: Variables: warn when a TargetConfig variable is shadowed by an environment variable
From `Pluma.create_target_context` method, warn when a target variable is shadowed by an environment variable when calling `variables.update()`

mjftw:
I added an extra commit to cut down some shorten some unit tests near the one I added too.
Not related, but may as well while I'm there.

Closes #97

### DIFF
--- a/pluma/cli/api.py
+++ b/pluma/cli/api.py
@@ -80,8 +80,8 @@ class Pluma:
 
         for variable, env_value in ((var, val) for var, val in env_vars.items()
                                     if var in context.variables):
-            log.warning(f'"{variable}" defined in environment variables and target config.'
-                        f'{os.linesep}Using environment: {env_value}')
+            log.warning(f'"{variable}" defined in both environment variables and target config.'
+                        f'{os.linesep}    Using environment: {env_value}')
 
         context.variables.update(env_vars)
         return context
--- a/pluma/cli/api.py
+++ b/pluma/cli/api.py
@@ -77,6 +77,12 @@ class Pluma:
         target_config = PlumaConfig.load_configuration('Target config', target_config_path,
                                                        PlumaConfigPreprocessor(env_vars))
         context = TargetConfig.create_context(target_config)
+
+        for variable, env_value in ((var, val) for var, val in env_vars.items()
+                                    if var in context.variables):
+            log.warning(f'"{variable}" defined in environment variables and target config.'
+                        f'{os.linesep}Using environment: {env_value}')
+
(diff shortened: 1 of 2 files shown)


pluma/cli/api.py
```diff
@@ -77,6 +77,12 @@ def create_target_context(target_config_path: str) -> PlumaContext:
         target_config = PlumaConfig.load_configuration('Target config', target_config_path,
                                                        PlumaConfigPreprocessor(env_vars))
         context = TargetConfig.create_context(target_config)
+
+        for variable, env_value in ((var, val) for var, val in env_vars.items()
+                                    if var in context.variables):
+            log.warning(f'"{variable}" defined in environment variables and target config.'
+                        f'{os.linesep}Using environment: {env_value}')
```
Pixep, line 84: Nit:
* defined in _both_ ...
* {os.linesep}    Using environment:
    * Using 4 spaces indentation here, as for other similar messages

tests/generic/cli/test_Pluma.py
```diff
@@ -37,57 +37,52 @@ def test_Pluma_create_target_context_should_parse_target_variables():
...
-    try:
-        os.environ['console'] = 'console'
-        os.environ['port'] = '/dev/random'
-        os.environ['some_env_var'] = 'echo "from env var"'
-        run_all('variable-env-sub-tests', 'variable-env-sub-target')
-    finally:
-        os.environ.clear()
-        os.environ.update(env)
+def test_Pluma_create_target_context_variables_should_reflect_env_vars(monkeypatch):
+    monkeypatch.setenv('abc', 'def')
```
Pixep, line 41: Much cleaner!

Pixep (changes requested):
Minor change regarding the output

pluma/cli/api.py
mjftw: I think I understood what you mean and have fixed, but let me know if I've misunderstood.

Pixep: I meant
> "{variable}" defined in 
to
> "{variable}" defined in **both**

and what we probably need 4 spaces before `Using environment:` if we want to match how similar messages are displayed :)

Pixep (approved):
Added that one comment, otherwise g2g

pluma/cli/api.py
mjftw: I think that's what I've done

PR #128, opened 2020-10-06
Rpi tests are now skipped off rpi rather than needing to be avoided

mjftw:
Check the output in the CI to see this in action. You should see these tests are now marked as SKIPPED, but are actually being read. This removes the need for only running the "generic" tests.

This also means you can now run the tests from the project root directory without the need for the `run_tests.sh` script.
```
python3 -m pytest tests
```


Pixep (approved):
Great, finally vscode will be happy out of the box :+1:

PR #131, opened 2020-10-06
tests: fix a few warnings in test_filesystem_plugins

Pixep:


mjftw (approved):
TBH I make little effort to stick to the line length rules with unit test function names as getting enough detail in the names is more important IMO.
Since the warnings complaining are likely to get us to try and keep the names short, do you think we could turn off the line length warnings for this?

PR #132, opened 2020-10-06
tests: remove mocks classes

Pixep:
Not used or needed anymore


Pixep:
@mjftw Couldn't see any use to these classes anymore, and the best code, is code we don't have of course :smiley_cat: . Should help have a more accurate coverage as well. Let me know what you think.

mjftw (approved):
Yep completely agree. These were added way back when we didn't have proper unit testing set up, and used for manual testing.
Not really required anymore, and zero code is 100% tested code ;)

`HardwareMock` and `BoardMock` may be useful for testing later on, but I suppose we could always retrieve them from the history. `BoardMock` especially, as it gives you a fake embeddd device with realistic console output.

PR #133, opened 2020-10-07
Create bump-version.yml

### LINKED ISSUE #129: Automatically bump version tag
**Is your feature request related to a problem? Please describe.**
Currently we have no version information for most of the history, and we are very far ahead of version `1.0.0` now.

**Describe the solution you'd like**
It would be good to remedy this by having the build version increment on every push to master.
We could use this Github action:
https://github.com/marketplace/actions/github-tag-bump

mjftw:
Closes #129.

This action will bump the revision/build/patch number on each push to master.
Once we've got this, each point in master will have an independent version number.
This also means we can set up CI to automatically push to Dockerhub and PyPi with the new versions as well.

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -1,0 +1,30 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Version tag bump
+
+# Controls when the action will run. Triggers the workflow on push or pull request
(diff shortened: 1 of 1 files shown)


Pixep (approved):
Cool, and its only updating the patch version.

Issue #96, opened 2020-09-25
Show additional test info when running "pluma check/list" 


mjftw:
**Is your feature request related to a problem? Please describe.**
As an enhancement to `pluma test` I'd like to get some info alongside the available tests list. 

Current:
```
pluma --plugin ./myplugindir check

Device actions:
    [x] {'copytodevice'}
          {"file": "/home/merlin/projects/pluma-automation/README.md", "destination": "/tmp/", "timeout": 10}

Core tests (testsuite, Python):
    [x] mytests.SayHello
    [ ] testsuite.kernel.KernelModulesLoaded
    [ ] testsuite.memory.MemorySize
    [ ] testsuite.networking.IperfBandwidth
    [ ] testsuite.networking.NetworkingTestBase
    [ ] testsuite.networking.RespondsToPing
```

**Describe the solution you'd like**
I'd like to see something like the following, with an entirely new sub-command `pluma list`:

```
pluma --plugin ./myplugindir list

Device actions:
    [x] {'copytodevice'}
          {"file": "/home/merlin/projects/pluma-automation/README.md", "destination": "/tmp/", "timeout": 10}

Core tests (testsuite, Python):
    [x] mytests.SayHello
            Print a hello message
    [ ] testsuite.kernel.KernelModulesLoaded
            Check whether certain kernel modules are installed
    [ ] testsuite.memory.MemorySize
            Check the memory of the target
    [ ] testsuite.networking.IperfBandwidth
    [ ] testsuite.networking.NetworkingTestBase
    [ ] testsuite.networking.RespondsToPing
```

This info could easily be pulled from the Test docstrings, and this would encourage people to actually write docstrings.
Maybe just the first line of the docstring here.

To take this one step further, it would be good to be able to query individual tests to see what there parameters are.
We'd display the full docstring here.

```
pluma list mytests

Core tests (testsuite, Python):
    [x] mytests.SayHello
            Print a hello message
    [ ] mytests.SomeOtherTest
            Really amazing docstring
```

```
pluma list mytests.SayHello
mytests.SayHello: selected
        Print a hello message
        Here's some other info that I put in the docstrings
        And some more!

Parameters:
        name (str)
        how_many_times (int)
```

The types for parameters could be pulled from the type hints in the `__init__` function.
It would be great if we could add a description of each parameter, but not sure there's a natural place to put this in the class without having to do an ugly docstring or something.

[...]

PR #137, opened 2020-10-08
testconfig: added test descriptions to 'tests' printout

Zematus:
Adds test-level descriptions for #96. Later will add parameter info for test-specific queries


mjftw:
For reference, this is the new console output with this PR:
```
(venv) merlin@pangolin:~/projects/pluma-automation$ pluma tests -c pluma.yml.sample -t pluma-target.yml.sample 

Device actions:
    [x] set
          {"device_console": "serial"}
          Sets the type of console to use
    [x] set
          {"device_console": "ssh"}
          Sets the type of console to use
    [x] wait
          {"duration": 2}
          Delays the execution of the next test/action for a set amount of time

Core tests (testsuite, Python):
    [ ] testsuite.filesystem.CheckFileSize
          No description
    [ ] testsuite.filesystem.FileExists
          No description
    [ ] testsuite.filesystem.FileIsBlockDevice
          No description
    [ ] testsuite.filesystem.FileIsCharDevice
          No description
    [ ] testsuite.filesystem.FileIsDir
          No description
    [ ] testsuite.filesystem.FileIsEmpty
          No description
    [ ] testsuite.filesystem.FileIsExecutable
          No description
    [ ] testsuite.filesystem.FileIsNotEmpty
          No description
    [ ] testsuite.filesystem.FileIsReadable
          No description
    [ ] testsuite.filesystem.FileIsRegular
          No description
    [ ] testsuite.filesystem.FileIsSocket
          No description
    [ ] testsuite.filesystem.FileIsSymlink
          No description
    [ ] testsuite.filesystem.FileIsWritable
          No description
    [x] testsuite.kernel.KernelModulesLoaded
          {"modules": ["wlan", "galcore"]}
          Verifies that a set of kernel modules are loaded
    [x] testsuite.memory.MemorySize
          {"total_mb": 985, "available_mb": 500}
          Tests the device's available and total memory
    [x] testsuite.networking.IperfBandwidth
          {"minimum_mbps": 10}
          Verifies the minimum bandwidth (in MBps)
    [x] testsuite.networking.RespondsToPing
          Verifies that there's a response to ping on the target

Inline tests (pluma.yml, Shell):
    [x] target_setup
          {"script": "echo \"setting things up on the target\"", "name": "target_setup"}
          Inline shell test
    [x] host_setup
          {"script": "echo \"setting things up locally\"", "run_on_host": true, "name": "host_setup"}
          Inline shell test
    [x] multiple_commands
          {"script": ["ls", "echo $USER"], "timeout": 2, "name": "multiple_commands"}
          Inline shell test
    [x] cat_meminfo
          {"script": "cat /proc/meminfo", "should_match_regex": ["MemFree:", "MemAvailable:"], "should_not_match_regex": ["error", "unknown"], "timeout": 2, "name": "cat_meminfo"}
          Inline shell test
    [x] cleanup
          {"script": "echo \"Cleaning up!\"", "name": "cleanup"}
          Inline shell test
```

.gitignore
```diff
@@ -107,3 +107,6 @@ venv.bak/
 
 # Pluma specific
 pluma*.yml
+
+#local development specific
+local-dev/
```
mjftw, line 112: Is this .gitignore definitely needed?

pluma/cli/testsconfig.py
```diff
@@ -199,6 +202,9 @@ def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = No
                     printed_data = json.dumps(test_parameters)
                     log.log(f'          {printed_data}', level=log_level)
 
+            if detail_level >= DetailLevel.MED:
+                log.log(f'          {test.description()}', level=log_level)
```
mjftw, line 206: It would be nice to log the test description in another colour so that it stands out more in the console output.
Maybe yellow?

Or maybe just inserting a newline after each test description would make the output easier to read.

pluma/cli/testsconfig.py
```diff
@@ -172,14 +172,17 @@ def selected_tests(self) -> List[TestDefinition]:
...
         return list(filter(lambda test: (test.selected), self.tests))
 
-    def print_tests(self, log_level: LogLevel = None):
-        TestsConfig.print_tests_definition(self.tests, log_level=log_level)
+    def print_tests(self, log_level: LogLevel = None, detail_level: DetailLevel = None):
+        TestsConfig.print_tests_definition(self.tests, log_level=log_level, detail_level=detail_level)
 
     @staticmethod
-    def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = None):
+    def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = None, detail_level: DetailLevel = None):
```
mjftw, line 179: Style: This line is a little over the 100char limit

pluma/cli/testsconfig.py
```diff
@@ -172,14 +172,17 @@ def selected_tests(self) -> List[TestDefinition]:
         '''Return only selected tests among all tests available'''
         return list(filter(lambda test: (test.selected), self.tests))
 
-    def print_tests(self, log_level: LogLevel = None):
-        TestsConfig.print_tests_definition(self.tests, log_level=log_level)
+    def print_tests(self, log_level: LogLevel = None, detail_level: DetailLevel = None):
+        TestsConfig.print_tests_definition(self.tests, log_level=log_level, detail_level=detail_level)
```
mjftw, line 176: Style: This line is a little over the 100char limit

pluma/core/baseclasses/logging.py
```diff
@@ -40,6 +40,12 @@
 }
 
 
+class DetailLevel(IntEnum):
```
mjftw, line 43: Is this idea of DetailLevel definitely needed?
Would simplify the code if we only had a description for each test, without detail level filtering.

pluma/test/shelltest.py
```diff
@@ -9,6 +9,7 @@
 
 
 class ShellTest(TestBase):
+    '''Inline shell test'''
```
mjftw, line 12: I think in this case I might actually remove the test description for this one.
Reason being that on the console output for `pluma tests` it's repeated a bunch and doesn't really add any extra information.

```
Inline tests (pluma.yml, Shell):
    [x] target_setup
          {"script": "echo \"setting things up on the target\"", "name": "target_setup"}
          Inline shell test
    [x] host_setup
          {"script": "echo \"setting things up locally\"", "run_on_host": true, "name": "host_setup"}
          Inline shell test
    [x] multiple_commands
          {"script": ["ls", "echo $USER"], "timeout": 2, "name": "multiple_commands"}
          Inline shell test
    [x] cat_meminfo
          {"script": "cat /proc/meminfo", "should_match_regex": ["MemFree:", "MemAvailable:"], "should_not_match_regex": ["error", "unknown"], "timeout": 2, "name": "cat_meminfo"}
          Inline shell test
    [x] cleanup
          {"script": "echo \"Cleaning up!\"", "name": "cleanup"}
          Inline shell test
```

What would be nice is if we let the user specify a description for the inline shell test in the YAML file.
E.g.

```
- shell_tests:
    target_setup:
      description: Sets up the target 
      script: echo "setting things up on the target"
    host_setup:
      description: Sets up the host
      script: echo "setting things up locally"
      run_on_host: true
``` 

These would then be used as the description output in `pluma test`

pluma/cli/config.py
```diff
@@ -89,6 +89,10 @@ def __init__(self, name: str, testclass: type, test_provider: object,
             raise ValueError(
                 f'Parameter sets for test "{name}" should be a list of dictionaries')
 
+    def description(self):
+        desc = self.testclass.__doc__ or "No description"
```
mjftw, line 93: Not sure if a test with no description should have it set to "No description", or just not have a description line printed in `pluma test`?
The advantage of "No description" is that it calls the user out for not putting a description, making it more likely that they will, but it doesn't tell them how.

I like the way that pytest handles this same thing with it's fixture view, calling out the specific line that the docstring is missing from, and highlighting it in red.

[See image here](https://ibb.co/fY7JYbP)

mjftw (changes requested):
It looks loads better having this test description output!
Some comments for change requests, mostly around improving output readability etc.

pluma/cli/config.py
mjftw: This should do it:
```python
In [1]: from inspect import getsourcefile, getsourcelines

In [2]: from pluma.plugins.testsuite.filesystem import FileIsSocket

In [3]: print(f'{getsourcefile(FileIsSocket)}:{getsourcelines(FileIsSocket)[1]} - missing docstring')
/home/merlin/projects/pluma-automation/pluma/plugins/testsuite/filesystem.py:53 - missing docstring
```

pluma/cli/testsconfig.py
Pixep: Yellow is used for warnings, so I'm not sure? This would need to be before arguments however I think, instead of after.

mjftw: > Yellow is used for warnings, so I'm not sure? This would need to be before arguments however I think, instead of after.

Oh yeah good point.

.gitignore
Zematus: not really, I moved that to my exclude file now

pluma/core/baseclasses/logging.py
Zematus: after the discussion, I think not. I removed it

pluma/cli/testsconfig.py
Zematus: removed extra parameter

pluma/cli/testsconfig.py
Zematus: removed extra parameter

pluma/cli/testsconfig.py
Zematus: I made a change so that only those tests that lack a description show "No description" in yellow as sort of a warning

pluma/test/shelltest.py
Zematus: I was trying to implement something like this, but I though that maybe it's better to change the base description to "Execute given script within the target (or host) shell" because that describes exactly what the test will do. Besides, adding a description parameter would make the parameter list also display the description as an extra test parameter

mjftw: > I was trying to implement something like this, but I though that maybe it's better to change the base description to "Execute given script within the target (or host) shell" because that describes exactly what the test will do. Besides, adding a description parameter would make the parameter list also display the description as an extra test parameter

Yeah fair point. As much as it would be nice, I'm not sure it's needed, and it's not simple to implement (where do we put that extra info?).

We could still display the parameters and their types, since we have that information already.
That coupled with a good description should let a user know if the test does what they're after or not.

pluma/test/shelltest.py
```diff
@@ -9,7 +9,7 @@
 
 
 class ShellTest(TestBase):
-    '''Inline shell test'''
+    '''Execute given script within the target (or host) shell'''
```
mjftw, line 12: Really minor one, but the sentence would read just as well with "given" removed.

mjftw:
For reference, how it currently looks. (we lose the colours here of course).

I think it's looking pretty good!
I like the newline between each of the test printouts, much easier to read.
The only thing I would say is that having the parameters for enabled tests and the descriptions together like that looks a bit confusing. 

```
(venv) merlin@pangolin:~/projects/pluma-automation$ pluma -c pluma.yml.sample -t pluma-target.yml tests

Device actions:
    [x] set
          {"device_console": "serial"}
          Sets the type of console to use

    [x] set
          {"device_console": "ssh"}
          Sets the type of console to use

    [x] wait
          {"duration": 2}
          Delays the execution of the next test/action for a set amount of time


Core tests (testsuite, Python):
    [ ] testsuite.filesystem.CheckFileSize
          No description

    [ ] testsuite.filesystem.FileExists
          No description

    [ ] testsuite.filesystem.FileIsBlockDevice
          No description

    [ ] testsuite.filesystem.FileIsCharDevice
          No description

    [ ] testsuite.filesystem.FileIsDir
          No description

    [ ] testsuite.filesystem.FileIsEmpty
          No description

    [ ] testsuite.filesystem.FileIsExecutable
          No description

    [ ] testsuite.filesystem.FileIsNotEmpty
          No description

    [ ] testsuite.filesystem.FileIsReadable
          No description

    [ ] testsuite.filesystem.FileIsRegular
          No description

    [ ] testsuite.filesystem.FileIsSocket
          No description

    [ ] testsuite.filesystem.FileIsSymlink
          No description

    [ ] testsuite.filesystem.FileIsWritable
          No description

    [x] testsuite.kernel.KernelModulesLoaded
          {"modules": ["wlan", "galcore"]}
          Verifies that a set of kernel modules are loaded

    [x] testsuite.memory.MemorySize
          {"total_mb": 985, "available_mb": 500}
          Tests the device's available and total memory

    [x] testsuite.networking.IperfBandwidth
          {"minimum_mbps": 10}
          Verifies the minimum bandwidth (in MBps)

    [x] testsuite.networking.NetworkingTestBase
          No description

    [x] testsuite.networking.RespondsToPing
          Verifies that there's a response to ping on the target


Inline tests (pluma.yml, Shell):
    [x] target_setup
          {"script": "echo \"setting things up on the target\"", "name": "target_setup"}
          Execute given script within the target (or host) shell

    [x] host_setup
          {"script": "echo \"setting things up locally\"", "run_on_host": true, "name": "host_setup"}
          Execute given script within the target (or host) shell

    [x] multiple_commands
          {"script": ["ls", "echo $USER"], "timeout": 2, "name": "multiple_commands"}
          Execute given script within the target (or host) shell

    [x] cat_meminfo
          {"script": "cat /proc/meminfo", "should_match_regex": ["MemFree:", "MemAvailable:"], "should_not_match_regex": ["error", "unknown"], "timeout": 2, "name": "cat_meminfo"}
          Execute given script within the target (or host) shell

    [x] cleanup
          {"script": "echo \"Cleaning up!\"", "name": "cleanup"}
          Execute given script within the target (or host) shell
```


One thing we could do for this is (cut down version for clarity):
```
(venv) merlin@pangolin:~/projects/pluma-automation$ pluma -c pluma.yml.sample -t pluma-target.yml tests

Device actions:
    [x] wait
          Delays the execution of the next test/action for a set amount of time
          Parameters:
               - duration: 2


Core tests (testsuite, Python):
    [ ] testsuite.filesystem.CheckFileSize
          No description - missing docstring
          Parameters:
              - path

    [ ] testsuite.filesystem.FileExists
          No description  - missing docstring
          Parameters:
              - path

Inline tests (pluma.yml, Shell):

    [x] multiple_commands
          Execute given script within the target (or host) shell
          Parameters:
              - script: ["ls", "echo $USER"]
              - timeout: 2
              - name: "multiple_commands"
          
    [x] cat_meminfo
          Execute given script within the target (or host) shell
          Parameters: 
              - script: "cat /proc/meminfo"
              - should_match_regex: ["MemFree:", "MemAvailable:"]
              - should_not_match_regex: ["error", "unknown"]
             -  timeout: 2
             -  name: "cat_meminfo"
```

Could even split it out into required and optional parameters, not sure.
I've also added " - missing docstring" to the no description bit, just so that users have a hint at how to fix the warning.

Pixep (changes requested):

pluma/test/shelltest.py
prototypicalpro: Fixed with 98b52e0486a34873de6186dfbabc919ec70cf94f.

pluma/cli/config.py
prototypicalpro: I added your code in 7d2cf078cadf94ee102358f01987140ee531e452. The output for a missing description now looks like this:
```console
[x] testsuite.networking.NetworkingTestBase
          No description - missing docstring at pluma/plugins/testsuite/networking.py:14 in NetworkingTestBase
```

pluma/utils/helpers.py
```diff
@@ -1,8 +1,15 @@
+from os.path import relpath
 import subprocess
 import json
 import re
 from datetime import datetime
+import os
+import sys
+import inspect
+from typing import Tuple, Optional, Type, Any
```
Pixep, line 9: `import` first, `from` then :)

pluma/utils/helpers.py
```diff
@@ -1,8 +1,15 @@
...
 
+_CUR_DIRECTORY = os.path.dirname(os.path.abspath(__file__))
+ROOT_PROJECT_DIRECTORY = os.path.normpath(os.path.join(_CUR_DIRECTORY, '../..'))
```
Pixep, line 12: You may want to use the more recent `Path` class instead, with `Path.cwd()` and friends

pluma/utils/helpers.py
```diff
@@ -66,3 +73,16 @@ def regex_filter_list(patterns, items, unique=None):
...
+    '''
+    Get the filename and line number of a given
+    python object type (@obj). Useful for creating a stack trace.
+    The filename will be relative to @relative_to for easy printing.
+    The default value of @relative_to is the project root of this library.
+    '''
+    src_file = inspect.getsourcefile(obj)
+    if src_file is None:
+        return None
+    rel_path = os.path.relpath(src_file, relative_to)
```
Pixep, line 87: Same about using the `Path` class

pluma/cli/testsconfig.py
```diff
@@ -196,13 +197,13 @@ def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = No
                     printed_data = json.dumps(test_parameters)
                     log.log(f'          {printed_data}', level=log_level)
 
-            if log_level == LogLevel.IMPORTANT:
```
Pixep: I _think_ you need to keep the LogLevel here, its used to only display the description when that content is important. Did you have any reason to remove it?

pluma/cli/testsconfig.py
```diff
@@ -174,7 +175,7 @@ def print_tests(self, log_level: LogLevel = None):
 
     @staticmethod
     def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = None):
-        if not log_level:
+        if log_level is None:
```
Pixep, line 178: Mh... I would argue here that `not value` is easier to read, and _I think_ the pythonic way of doing it. Less strict of course, but I don't know that it matters. Thoughts @mjftw ?

pluma/utils/helpers.py
```diff
@@ -66,3 +73,16 @@ def regex_filter_list(patterns, items, unique=None):
     if unique:
         gen = set(gen)
     return sorted(gen)
+
+def get_file_and_line(obj: Type[Any], relative_to: str = ROOT_PROJECT_DIRECTORY) -> Optional[Tuple[str, int]]:
```
Pixep, line 77: Because we return a Tuple here, it might be safer to return a None Tuple instead of Optional, otherwise it will screw up code that is using it, which is unlikely to test for the return value. What do you think?

pluma/cli/config.py
```diff
@@ -92,6 +92,11 @@ def __init__(self, name: str, testclass: type, test_provider: object,
     def __repr__(self):
         return f'{self.__module__}.{self.__class__.__name__}{self.parameter_sets or ""}'
 
+    def description(self):
```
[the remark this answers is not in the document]
Pixep: @prototypicalpro Could you incorporate these comments as well? Thank you :)

pluma/test/testbase.py
```diff
@@ -35,3 +35,7 @@ def save_data(self, data: dict = None, **data_kwargs: dict):
     def __repr__(self):
         """Return a human-readable name for the test"""
         return self._test_name
+
+    def description(self):
```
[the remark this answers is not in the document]
Pixep: @prototypicalpro Could you incorporate these comments as well? Thank you :)

Pixep (changes requested):

pluma/cli/testsconfig.py
prototypicalpro: In this specific instance one possible value of `log_level` is `LogLevel.DEBUG = 0`, which is falsey and would create an unintended feature with the standard pythonic if statement.

pluma/test/testbase.py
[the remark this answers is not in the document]
prototypicalpro: It looks like you can't combine classmethod and property (lame), so I'll have to stick with just a classmethod.

pluma/cli/testsconfig.py
prototypicalpro: Is there a reason why that is switching off of the LogLevel, as opposed to a boolean parameter? If we're not sure I can leave it in.

pluma/test/shelltest.py
prototypicalpro: Resolved with https://github.com/Witekio/pluma-automation/pull/137/commits/9c65fe5b6ded20d23c29d27afd9b4d01c75c667d where the test configuration is shown in YAML with the test description.

pluma/cli/testsconfig.py
prototypicalpro: I went ahead and did some digging, and fount two places where this statement mattered. The first one is in `execute_tests`: https://github.com/Witekio/pluma-automation/blob/0e26c783b80a37c14a6bff88177853ccdf1c908a/pluma/cli/api.py#L56-L64 and the second one is in `build_test_controller`: https://github.com/Witekio/pluma-automation/blob/0e26c783b80a37c14a6bff88177853ccdf1c908a/pluma/cli/api.py#L117-L123

It looks to me (feel free to correct me) that the intention was to have a verbose output with `pluma tests` and a less verbose output with `pluma check`, and this behavior was switched using the `log_level` parameter. Since these are the only two places where `print_tests_definition` is called, I went ahead and refactored this functionality into a `show_description` keyword argument in the print tests function to make this behavior more clear.

pluma/cli/api.py
```diff
@@ -61,7 +61,7 @@ def execute_tests(tests_config_path: str, target_config_path: str):
 
         log.log(
             'List of core and script tests available, based on the current configuration.')
-        tests_config.print_tests(log_level=LogLevel.IMPORTANT)
+        tests_config.print_tests(log_level=LogLevel.IMPORTANT, show_description=True)
```
Pixep, line 64: Okay nice. Man I get confused everytime with the name of the method... its a bad name. Its the name of the CLI command, `tests`, but... nah. Anyway, yes that works 👍

Pixep (approved):
Good for me. @mjftw ?

pluma/cli/api.py
mjftw: Nice solution with clearer intent.

pluma/cli/testsconfig.py
```diff
@@ -181,18 +183,34 @@ def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = No
...
                     color='green' if test.selected else 'normal', level=log_level)
 
-            for test_parameters in test.parameter_sets:
-                if test.selected and test_parameters:
-                    printed_data = json.dumps(test_parameters)
-                    log.log(f'          {printed_data}', level=log_level)
+            if show_description:
+                description = test.description
+                if description is not None:
+                    log.log(f'          {description}', level=log_level)
```
mjftw, line 199: Very picky point here, and I'm not going to block merging on this.

Would be better to not have an explicit load of spaces here before every bit of info to be printed.
I believe the intent is to print with indentation at different levels, and if this is the case we could do something like:

```python
indent = '     '

log.log(f'{indent}Message with indentation')
log.log(f'{indent*2}Message with double indentation')
log.log(f'{indent*3}Message with triple indentation')
```
_______________________________________________

Or if this is a general thing that we may want to do all over the place when we log (and I think it is), we could put this code inside the logger and have an indentation parameter for the log functions.

E.g.
```python
log.log(f'{description}', level=log_level, indent=2)
```

This is of course a more complicated solution though, and would require us to create (via inheritance) a modified version of the standard Python logger to use. Complexity is a bad thing, so maybe this idea is too.

pluma/cli/testsconfig.py
```diff
@@ -181,18 +183,34 @@ def print_tests_definition(tests: List[TestDefinition], log_level: LogLevel = No
         last_provider = None
         for test in tests:
             if test.provider != last_provider:
-                log.log(f'{os.linesep}{test.provider.display_name()}:',
+                if last_provider is not None:
+                    log.log('', level=log_level)
```
mjftw, line 187: What is this empty log for?

mjftw (approved):
This is looking really good!

My only comment is really very minor so approving; feel free to merge.

pluma/cli/testsconfig.py
prototypicalpro: To add a trailing newline:
```console
    [x] testsuite.networking.NetworkingTestBase
    [x] testsuite.networking.RespondsToPing
         < --- this one
Inline tests (pluma.yml, Shell):
    [x] target_setup
```

pluma/cli/testsconfig.py
prototypicalpro: Would https://github.com/Witekio/pluma-automation/pull/137/commits/4a8924be9cdb51dab59b1d84f592930fdf1e39d5 be what you are looking for? I'm not sure what you are referring to regarding the modified version of the python logger.

mjftw: Oh yeah perfect, that's exactly what I was after!

mjftw (approved):
This is looking great!

A few things to tidy up as found in the Codacy Static Code Analysis (CI step below), then good to merge 👍🏼 
https://app.codacy.com/gh/Witekio/pluma-automation/pullRequest?prid=6303685

Pixep:
Good to go, only static type check failed, as expected.

PR #146, opened 2020-10-10
tests: report coverage even if some test fail

Pixep:


Pixep:
Quick fix to coverage report on master

PR #148, opened 2020-10-10
Remove legacy asynchronism classes

Pixep:
@mjftw Spotted these ones as uncovered and unused. Do you want to keep them or remove them?

There is also the `nonblocking` function in utils.py, which is not affected by this PR.


mjftw:
The reason for AsyncSampler is that it is used any time the multimeter is used in order to allow it to sample whole other stuff is going on. It's not directly depended upon because didn't seem right to do, but it is useful.

Nonblocking is used as a Mixin for any classes that need to run async, allowing individual methods to be marked as async. When run they do so in a separate thread and the results can be collected later.
A better way of doing the same thing would be using asyncio and async await, but that requires more effort with setting up async loops which isn't really worth it in this case.
It is currently unused though yes as I never got round to adding it to anything. Is a good candidate for a bunch of the hardware control classes that need to run async (e.g. Modem & Multimeter Classes)

TL;DR
I think they're still useful and worth keeping around.

Pixep:
Alrighty! Maybe I should let you do some cleanup in legacy classes @mjftw ? Usually, I look at the worst coverage files from Codacy as a starting point.

PR #150, opened 2020-10-10
tests: move SerialConsole tests over to PexpectEngine/ConsoleBase

Pixep:
Will help mitigate CI tests errors, and we are actually unit testing the right classes


tests/generic/utils.py
```diff
@@ -9,9 +9,9 @@ def fd_has_data(fd, timeout=0):
 
 
 class OsFile:
-    def __init__(self, fd, encoding):
+    def __init__(self, fd, encoding=None):
         self.fd = fd
-        self.encoding = encoding
+        self.encoding = encoding or 'ascii'
```
mjftw, line 14: We might want to use UTF-8 here instead?

mjftw (approved):
This looks great!
Looks like we're heading towards not having any of those annoying false CI failures!

Only a minor comment, but approving anyway as it really is minor.

pluma/core/baseclasses/pexpectengine.py
```diff
@@ -32,7 +32,7 @@ def _on_opened(self):
 
     @property
     def is_open(self):
-        return self._pex and self._pex.isalive()
+        return (self._pex and self._pex.isalive()) is True
```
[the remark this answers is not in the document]
Pixep: Got it, thanks. Done.

tests/generic/utils.py
Pixep: Probably yes, thought console engine default to ascii

PR #152, opened 2020-10-12
Fix/various3

Pixep:
Various line length and warnings fixes


pluma/core/smartplug.py
```diff
@@ -1,2 +0,0 @@
-# class KasaSmartPlug:
```
mjftw, line 1: Oops! How did this sneak in? (definitely my doing)

mjftw (approved):
Looks good!

PR #154, opened 2020-10-12
Feature/refactor testrunner

mjftw:
Lots of changes to TestRunner, and lots more tests.
I've also made `board` optional in both `TestRunner` and `TestBase`, since it's not actually really needed; should make classes more flexible.
I also found and fixed a bug along the way, but I may have caused it in my refactoring, not sure.


pluma/test/testrunner.py
```diff
@@ -262,25 +210,31 @@ def _run_task(self, task_name, test_name):
...
+            if not self.continue_on_fail:
+                raise e
+
         else:
-            self.board.log('PASS', color='green', level=LogLevel.IMPORTANT, bypass_hold=True)
+            self.log('PASS', color='green', level=LogLevel.IMPORTANT, bypass_hold=True)
         finally:
-            self.board.release_log()
+            if self.board:
+                self.board.release_log()
```
Pixep, line 234: `release_log` should occur regardless of if you have a board or not, and probably be `self.release_log()` actually :). Probably a miss. I don't know if its the right time to move to `log.log()` functions, as you are changing those?

Pixep:
Great, happy we can clean all of those for a fresh start. Only 1 check ran, and I don't know if the target branch if the right one (?). Will review a bit more in depth, but looks pretty cool

mjftw:
> Great, happy we can clean all of those for a fresh start. Only 1 check ran, and I don't know if the target branch if the right one (?). Will review a bit more in depth, but looks pretty cool

Fixed it, looks like the checks have run now 👍🏼

pluma/test/testrunner.py
```diff
@@ -64,48 +94,17 @@ def _init_test_data(self, test):
...
 
-        # Check if any tasks failed
-        if self.test_fails:
-            return False
-        else:
-            return True
+    @tests.setter
+    def tests(self, tests: List[TestBase]):
+        for test in tests:
+            self.add_test(test)
```
Pixep, line 104: So doing `tests = x` if effectively doing something like `tests += x` actually right? Is that wanted?

pluma/test/testrunner.py
```diff
@@ -240,17 +191,14 @@ def _run_task(self, task_name, test_name):
...
         column_limit = 75
         if len(test_message) > column_limit:
             test_message = f'{test_message[:column_limit-3]}...'
 
-        self.board.log(test_message.ljust(column_limit) + ' ',
+        self.log(test_message.ljust(column_limit) + ' ',
                        level=LogLevel.IMPORTANT, newline=False)
-        self.board.hold_log()
+        if self.board:
+            self.board.hold_log()
```
Pixep, line 201: Same comment as `release_log`

pluma/test/testrunner.py
mjftw: Ooooooooooops 👍🏼. Fixed.

pluma/test/testrunner.py
mjftw: Fixed

pluma/test/testrunner.py
```diff
@@ -262,25 +219,30 @@ def _run_task(self, task_name, test_name):
...
+            self._failed_task_side_effects(test, task_name, e)
+
+            if not self.continue_on_fail:
+                raise e
+
         else:
-            self.board.log('PASS', color='green', level=LogLevel.IMPORTANT, bypass_hold=True)
+            self.log('PASS', color='green', level=LogLevel.IMPORTANT, bypass_hold=True)
         finally:
             self.board.release_log()
```
Pixep, line 242: `self.release_log()` here if you want to match the `self.hold_log()` you have above

pluma/test/testrunner.py
```diff
@@ -262,25 +219,30 @@ def _run_task(self, task_name, test_name):
...
 
-    def _handle_failed_task(self, test, task_name, exception, abort=True):
+    def _failed_task_side_effects(self, test: TestBase, task_name: str, exception: Exception):
```
Pixep, line 244: nit: I actually maybe preferred the older name, which was a verb. I guess this could be `_on_task_failed` or `_run_task_failed_side_effects`, or the original too. This one could be read as just "get" the side effects. But that's not super important.

pluma/test/testrunner.py
```diff
@@ -316,9 +276,30 @@ def send_fail_email(self, exception, test_failed, task_failed):
...
+    def _run(self, tests: Iterable[TestBase]):
+        self.log('== TESTING MODE: SEQUENTIAL ==', color='blue', bold=True,
+                 level=LogLevel.DEBUG)
+
+        for test in tests:
+            for task_name in self.known_tasks:
+                self._run_tasks(test, task_name)
+
+
+class TestRunnerParallel(TestRunnerBase):
```
Pixep, line 299: Out of curiosity, this one isn't used yet, right?

Pixep (changes requested):
Only one important change for `release_log`

pluma/test/testrunner.py
mjftw: Oops! Fixed

pluma/test/testrunner.py
mjftw: Yeah fair point, went with the old name

pluma/test/testrunner.py
mjftw: Nope, we could remove it entirely tbh

Pixep (approved):
Good to go!

mjftw:
> Good to go, once the pipeline is fixed :)
> 
> ```
> _______________ ERROR collecting generic/test/test_TestRunner.py _______________
> import file mismatch:
> imported module 'test_TestRunner' has this __file__ attribute:
>   /home/runner/work/pluma-automation/pluma-automation/tests/generic/core/test_TestRunner.py
> ```

I had a look and tried to work this out, but I'm really not sure what's causing it...
Doesn't happen on my machine

PR #156, opened 2020-10-13
tests: rename generic/generic to generic/core

Pixep:
Renamed to match pluma modules


mjftw (approved):
Aah okay to be honest I didn't actually know why we had `tests/generic/generic` alongside `tests/generic/cli`, and I've been putting all sorts of stuff in there!
Completely agree with this PR, will have to fix one of my PRs to put `test_TestRunner.py` in a different directory ;)

PR #159, opened 2020-10-13
Feature/speed up unit tests

mjftw:
Run tests in parallel for big speed gains!

@Pixep, please could you run the yocto build stuff on your machine to check I haven't broken anything?
I haven't got it set up on mine.


Pixep:
Fun, I was actually playing with running tests in parallel yesterday :)

Pixep:
Could you rebase on top of https://github.com/Witekio/pluma-automation/pull/160?
I changed the paths and retested all there, as well as added a bunch of unit tests.

mjftw:
> Could you rebase on top of #160?
> I changed the paths and retested all there, as well as added a bunch of unit tests.

Done

Pixep:
Thanks, the rebase apparently left some duplicated code that was move to a different `filebuilder.py` file (visible in Codacy duplications). And that code was merged onto master. Can you rebase once again, dropping your `
TestsBuilder: switch to using pathlib over os.path` commit?

... then hopefully we should be good! Thanks in advance :)

mjftw:
All checks passing, ready for review @Pixep  👍🏼

mjftw:
Finally fixed the duplicated code issue and this is ready to merge :)

Pixep:
@prototypicalpro Could you check the unit tests pass on your end too?

prototypicalpro:
The parallel running works great! My only note is I initially ran into a problem where `pytest-xdist` was not installed b/c I had run the install script on master branch and then ran the tests on this branch--to remedy this, I went ahead and added a check in `run_tests.sh` that will suggest the user re-run the install script if `pytest-xdist` is not installed.

prototypicalpro (approved):

prototypicalpro:
As a side note, a test failed when I pushed that commit, but succeeded after I manually re-ran it. Here's the failure log if it's useful:
<details>
  <summary>
  Test Log
  </summary>     

    _______ test_SerialConsole_check_alive_returns_true_when_target_responds _______
    
    serial_console_proxy = <conftest.SerialConsoleProxy object at 0x7f3bc7caff98>
    
        def test_SerialConsole_check_alive_returns_true_when_target_responds(serial_console_proxy):
            async_result = nonblocking(
                serial_console_proxy.console.check_alive)
        
            # Send console a newline char
            serial_console_proxy.fake_reception(serial_console_proxy.console.engine.linesep)
        
    >       assert async_result.get() is True
    E       assert False is True
    E        +  where False = <bound method ApplyResult.get of <multiprocessing.pool.ApplyResult object at 0x7f3bc79c59e8>>()
    E        +    where <bound method ApplyResult.get of <multiprocessing.pool.ApplyResult object at 0x7f3bc79c59e8>> = <multiprocessing.pool.ApplyResult object at 0x7f3bc79c59e8>.get
    
    tests/generic/core/test_SerialConsole.py:33: AssertionError
    ----------------------------- Captured stdout call -----------------------------
    msg b'\n'
</details>

Pixep:
> The parallel running works great! My only note is I initially ran into a problem where `pytest-xdist` was not installed b/c I had run the install script on master branch and then ran the tests on this branch--to remedy this, I went ahead and added a check in `run_tests.sh` that will suggest the user re-run the install script if `pytest-xdist` is not installed.

Sounds cool, thank you Noah! Usually we don't commit on others' branches, but I'm sure @mjftw will be fine with it. I let you confirm if its all good @mjftw :).

> As a side note, a test failed when I pushed that commit, but succeeded after I manually re-ran it. Here's the failure log if it's useful:
Yes you are right, this is one of those flaky tests, which is pretty hard to fix for the CI, for reasons partly unknowns, but just works locally. Made some moves to make it a little better, there's likely a problem on the CI environment with files and virtual TTYs.

mjftw:
> The parallel running works great! My only note is I initially ran into a problem where `pytest-xdist` was not installed b/c I had run the install script on master branch and then ran the tests on this branch--to remedy this, I went ahead and added a check in `run_tests.sh` that will suggest the user re-run the install script if `pytest-xdist` is not installed.

Ah okay yeah I can see why that would happen, good spot.

Might I suggest that rather than putting in a special case check asking the user to take action, that we just make it so that the parallel test execution only happens _if_ the pytest-xdist package is installed? Otherwise we can just continue to run the tests sequentially. After all, the parallel testing is only an optimisation, and not a real requirement to running the tests.

I'm a little hesitant to put in specific code for this special case as it would only ever manifest itself to developers, moving from a previous revision (where pytest-xdist was in the project dependencies) to this one, and never occur for a real user of the tool. 

No problem about pushing to this branch, and normally I'd say let's just leave the commit in, but in this case I may go ahead and revert it. That is, if you agree with my alternative solution, I'll implement that instead.

Thanks for the contribution though 👍🏼

mjftw:
> > As a side note, a test failed when I pushed that commit, but succeeded after I manually re-ran it. Here's the failure log if it's useful:
> 
> Yes you are right, this is one of those flaky tests, which is pretty hard to fix for the CI, for reasons partly unknowns, but just works locally. Made some moves to make it a little better, there's likely a problem on the CI environment with files and virtual TTYs.

Yeah... we really need to fix that somehow.
Feels like us getting into the expectation that one of the tests might fail and that's okay is likely to lead to other genuine failures being missed.
See [Broken Window Theory](https://blog.codinghorror.com/the-broken-window-theory/)

prototypicalpro:
>This is a good fix for the issue you came across, but maybe we don't need to fix it at all.
Might I suggest that rather than putting in a special case check asking the user to take action, that we just make it so that the parallel test execution only happens if the pytest-xdist package is installed? Otherwise we can just continue to run the tests sequentially. After all, the parallel testing is only an optimization, and not a real requirement for running the tests.

That sounds like a good approach to me👍

Pixep:
> > This is a good fix for the issue you came across, but maybe we don't need to fix it at all.
> > Might I suggest that rather than putting in a special case check asking the user to take action, that we just make it so that the parallel test execution only happens if the pytest-xdist package is installed? Otherwise we can just continue to run the tests sequentially. After all, the parallel testing is only an optimization, and not a real requirement for running the tests.
> 
> That sounds like a good approach to me👍

As you made that change, could you adjust it according to Merlin's comment, Noah? Amending your previous commit. Then this should be ready to be merged :)

prototypicalpro:
Done! It should be good to go🎉

Pixep (approved):

Pixep:
Great, thanks!

PR #160, opened 2020-10-16
Feature/improve test builder

Pixep:
Architectural step towards supporting CMake (and other) builders:
* Added new `FileBuilder` base class
* Added new `CommandFileBuilder` base class
* Added test coverage
* Yocto C builder class now inherits `FileBuilder`, and leverages `CommandFileBuilder` to build files.


pluma/cli/testsbuilder.py
```diff
@@ -11,7 +15,66 @@ class TestsBuildError(Exception):
...
+        '''Generate the output file, and return its path.'''
+
+    @abstractmethod
+    def clean(self, force: bool = False):
+        '''Clean build and output files.'''
+
+    @property
+    def output_filepath(self):
+        '''Return output file path'''
+        return os.path.join(self.install_dir, self.target_name)
```
mjftw, line 36: Better to use more modern pathlib over `os.path` for new code:
https://docs.python.org/3/library/pathlib.html

pluma/cli/testsbuilder.py
```diff
@@ -11,7 +15,66 @@ class TestsBuildError(Exception):
...
+
+    @staticmethod
+    def create_directory(directory: str):
+        '''Create the directory or raise an error'''
+        try:
+            os.makedirs(directory, exist_ok=True)
```
mjftw, line 42: pathlib can do this too: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

pluma/cli/testsbuilder.py
```diff
@@ -11,7 +15,66 @@ class TestsBuildError(Exception):
     pass
 
 
-class TestsBuilder:
+class FileBuilder(ABC):
+    '''Generic builder class to generate an output file from inputs.'''
+
+    def __init__(self, target_name: str, install_dir: str = None):
+        self.target_name = target_name
+        self.install_dir = os.path.abspath(install_dir)
```
mjftw, line 23: pathlib can do this, although not sure where it's documented.
https://stackoverflow.com/questions/42513056/how-to-get-absolute-path-of-a-pathlib-path-object

pluma/cli/testsbuilder.py
```diff
@@ -11,7 +15,66 @@ class TestsBuildError(Exception):
...
+
+    def __init__(self, target_name: str, build_command: str, install_dir: str = None):
+        super().__init__(target_name, install_dir)
+        self.build_command = build_command
+
+    def build(self) -> str:
+        log.info(f'Build "{self.target_name}"...')
+        FileBuilder.create_directory(self.install_dir)
+
+        log.debug(f'Build command = {self.build_command}')
```
mjftw, line 59: May not be appropriate here but mentioning as it's a cool new feature in Python 3.8 that is related to this and is good to be aware of:

https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

pluma/cli/testsbuilder.py
```diff
@@ -24,16 +87,47 @@ class TestsBuilder:
...
     @staticmethod
-    def create_directory(directory):
+    def clean(force: bool = False):
         try:
-            os.makedirs(directory, exist_ok=True)
+            # TODO: Would need to use actual build folder and list of built/installed
+            # files
+            build_folder = YoctoCCrossCompiler.DEFAULT_BUILD_ROOT
+
+            if not force:
```
mjftw, line 110: Nice to ask :)

mjftw (changes requested):
Code all looks good. Definitely a step forward for architecture!
Only comments are super minor and are about using `pathlib` over `os.path`.

pluma/cli/testsbuilder.py
Pixep: Done

pluma/cli/testsbuilder.py
Pixep: Done

pluma/cli/testsbuilder.py
Pixep: Done. Went for `resolve()` that I found in the doc, and SO seems to confirm that's the wtg.

pluma/cli/testsbuilder.py
Pixep: Neat! We're compatible down to 3.6 though, so not really usable here.

mjftw (commented):
Good to go

mjftw (approved):
Missed pressing the Approve button on the last review!

PR #162, opened 2020-10-19
DeviceActions: use consistent naming for actions

Pixep:
Use '_', and disable default DeviceActionRegistry.register, to force
using a specific name.


mjftw (approved):
Much better 👍🏼

PR #163, opened 2020-10-19
DeployAction: support deploying multiple files

Pixep:
Support multiple files for "deploy" action, check that the console supports it, and add unit tests


mjftw (approved):
Nice one!

PR #165, opened 2020-11-04
Fix `reboot_and_validate` and tutorial examples

callumdavidobrien:


docs/tutorials/2-5-tutorial-board.md
```diff
@@ -123,6 +124,7 @@ If these defaults do not work for your system, they can be changed with the boar
 
 ```python
 from pluma import Board
+from pluma.core.dataclasses import SystemContext, Credentials
```
mjftw, line 127: Oops, this should be accessible with
```python
from pluma import SystemContext, Credentials
```
It's not, because we missed adding some imports to `pluma/__init__.py`
Since you're making a change here already, please could you add the import to the `__init__.py` file so that this line can be as above?

mjftw: Oh one thing to note is that everything from `pluma.core` is accessible just from `pluma`, as it's got a global import for the core in the pluma __init__, so this is actually a change to `pluma/core/__init__.py`.

callumdavidobrien: Yeah I saw that in `pluma/__init__.py` :slightly_smiling_face:

mjftw (approved):
Looks great, thanks for fixing that.
Merge away

Pixep:
Great thanks. That's a strong hint that this one needs to be covered :)

PR #166, opened 2020-11-06
Fix some issues with SoftPower in the CLI

callumdavidobrien:
- `SoftPower` couldn't be used through the CLI with a serial console as `TargetFactory.create_power_control` was only ever passed the SSH console;
- There was no way to specify the `on_cmd` and `off_cmd` for a `SoftPower` power control, so it couldn't actually be used to control anything.


mjftw (approved):
That's excellent, thanks for putting this in.

Sorry it took so long to get to reviewing this!

PR #171, opened 2020-12-28
Expose log file for SSH and Serial console in the target configuration

Pixep:
Pushing this waiting to be testing with an actual device. Unit tests added and passing.


mjftw (approved):
Looks great, nice one!

pluma/core/baseclasses/pexpectengine.py
```diff
@@ -15,20 +15,13 @@ def __init__(self, linesep: str = None, encoding: str = None, raw_logfile: str =
                          raw_logfile=raw_logfile)
         self._pex = None
 
-    def _open_process(self, command: str):
-        self._pex = pexpect.spawn(command, timeout=0.01)
-        self._on_opened()
+    def _open_process(self, command: str, log_file: IO = None):
```
Pixep, line 18: We now handle process (assuming SSH, ... ) and FD-based (assuming TTYs) differently, to prevent double echos in the log files. @mjftw any opinion on that?

pluma/core/baseclasses/pexpectengine.py
```diff
@@ -15,20 +15,13 @@ def __init__(self, linesep: str = None, encoding: str = None, raw_logfile: str =
...
+        self._pex = pexpect.spawn(command, timeout=0.01, logfile=log_file)
 
-    def _open_fd(self, fd):
+    def _open_fd(self, fd: int, log_file: IO = None):
         self._pex = pexpect.fdpexpect.fdspawn(fd=fd, timeout=0.5)
-        self._on_opened()
-
-    def _on_opened(self):
```
Pixep, line 26: No more _on_opened, you're going to be happy @mjftw ;)

pluma/core/baseclasses/pexpectengine.py
mjftw: I think that's okay, it's not a major difference and it will improve output readability

mjftw (approved):
Looking good. Still waiting to test on real hardware?

Pixep:
> Looking good. Still waiting to test on real hardware?

Right, did some first tests, but I'm not sure this is stable, I had contradicting results with the log, sometimes truncated...

pluma/core/baseclasses/consoleengine.py
```diff
@@ -36,41 +36,44 @@ def __init__(self, linesep: str = None, encoding: str = None,
...
+                self._open_fd(fd=console_fd, log_file=self._raw_logfile_io)
                 self._console_type = ConsoleType.FileDescriptor
 
             assert self.is_open
         except Exception:
             raise ConsoleCannotOpenError
 
     @abstractmethod
-    def _open_process(self, command: str):
+    def _open_process(self, command: str, log_file: IO = None):
```
prototypicalpro, line 72: It looks like the tests are failing b/c the `MockConsoleEngine` does not implement the `log_file` kwarg in it's `_open_process` and `_open_fd` methods. I was wondering how the abstract base class didn't catch this discrepency, and discovered that Python doesn't support validating keyword arguments in abstract methods: https://bugs.python.org/issue24183.  I can fix the `MockConsoleEngine` implementation, but it might be worth discussing if this keyword argument should be replaced with something that can be validated.

prototypicalpro:
It looks like one of the pexpect tests is flakey, I re-ran it and it passed. Here's the error log:
<details>
<summary>Error Log</summary>

    ______ test_PexpectEngine_wait_for_match_should_return_after_timeout[0.2] ______
    
    pty_pair = PtyPair(main=<utils.OsFile object at 0x7f687a475ee0>, secondary=<utils.OsFile object at 0x7f687a4755b0>)
    timeout = 0.2
    
        @pytest.mark.parametrize('timeout', [0.2, 1])
        def test_PexpectEngine_wait_for_match_should_return_after_timeout(pty_pair, timeout):
            engine = PexpectEngine()
            engine.open(console_fd=pty_pair.main.fd)
        
            start_time = time.time()
            engine.wait_for_match(match=['abc'], timeout=timeout)
        
    >       assert 0.8 * timeout < time.time() - start_time < 1.2*timeout
    E       assert (1609882225.0173304 - 1609882224.716025) < (1.2 * 0.2)
    E        +  where 1609882225.0173304 = <built-in function time>()
    E        +    where <built-in function time> = time.time
    
    tests/generic/core/test_PexpectEngine.py:131: AssertionError
</details>

pluma/core/baseclasses/consoleengine.py
Pixep: Good point, I saw that yesterday after pushing, will make a quick fix for it. What are you thinking of when you say:

> but it might be worth discussing if this keyword argument should be replaced with something that can be validated

Were you able to see if the rest works by chance? Meaning the serial and SSH log being properly (and fully) populated?

prototypicalpro: I was able to test the rest, and it seems to work!

> What are you thinking of when you say:
>> but it might be worth discussing if this keyword argument should be replaced with something that can be validated

Not sure to be honest. Since it's not that big of an issue it would probably be best to just leave it and come back if it causes any more problems.

prototypicalpro (approved):

Pixep:
TODO: Cover with a unit test

tests/generic/core/test_SerialConsole.py
```diff
@@ -314,3 +317,23 @@ def test_SerialConsole_login_except_on_wrong_username_match(serial_console_proxy
...
 def test_SerialConsole_does_not_require_login(serial_console_proxy):
     assert serial_console_proxy.console.requires_login is True
+
+
+def test_SerialConsole_raw_log_file(pty_pair):
+    received = 'abcd\nefgh'
+
+    with tempfile.NamedTemporaryFile(delete=False) as tmpfile:
+        console = SerialConsole(
+            port=os.ttyname(pty_pair.main.fd),
```
prototypicalpro, line 327: I was able to get this test to pass by switching the `main` and `secondary` fds:
```suggestion
            port=os.ttyname(pty_pair.secondary.fd),
```
I think this might have something to do with how the data is being moved from the `main` to the `secondary` fd and vice versa. Running the script below to write from the `main` fd to the `secondary` fd generates the following output:
```python
main, secondary = pty.openpty()
os.write(main, b'hello world!\n')
print(os.read(secondary, 13))
```
```console
b'hello world!\n'
```
If we reverse `main` and `secondary`, we get a similar but slightly different result:
```python
main, secondary = pty.openpty()
os.write(secondary, b'hello world!\n')
print(os.read(main, 13))
```
```console
b'hello world!\r'
```

I think that this transformation when writing from `secondary` to `main` (or something else weird about the PTY) is causing the `Serial` class to drop any data it receives when reading from the `main` fd, shown below:
```python
main, secondary = pty.openpty()
serial = Serial(os.ttyname(main), timeout=1)
os.write(secondary, b'hello world!\n')
print(serial.read(13))
```
```console
b''
```
```python
main, secondary = pty.openpty()
serial = Serial(os.ttyname(secondary), timeout=1)
os.write(main, b'hello world!\n')
print(serial.read(13))
```
```console
b'hello world!\n'
```
I'm not familiar enough with linux to know what could be causing this behavior with the PTY, but it's certainly very strange.

tests/generic/core/test_SerialConsole.py
```diff
@@ -314,3 +317,23 @@ def test_SerialConsole_login_except_on_wrong_username_match(serial_console_proxy
 
 def test_SerialConsole_does_not_require_login(serial_console_proxy):
     assert serial_console_proxy.console.requires_login is True
+
+
+def test_SerialConsole_raw_log_file(pty_pair):
+    received = 'abcd\nefgh'
+
+    with tempfile.NamedTemporaryFile(delete=False) as tmpfile:
```
prototypicalpro, line 325: I don't think you need to keep the file around outside the test, feel free to correct me if I'm wrong.
```suggestion
    with tempfile.NamedTemporaryFile() as tmpfile:
```

tests/generic/core/test_SerialConsole.py
```diff
@@ -314,3 +317,23 @@ def test_SerialConsole_login_except_on_wrong_username_match(serial_console_proxy
...
+        # tmpfile.name
+
+        pty_pair.secondary.write(received)
+        time.sleep(1)
+        print(console.read_all())
+        time.sleep(1)
+        console.close()
+        time.sleep(1)
+
+        assert tmpfile.read() == received
```
prototypicalpro, line 339: decode `bytes` to `str`.
```suggestion
        assert tmpfile.read().decode() == received
```

tests/generic/core/test_SerialConsole.py
```diff
@@ -314,3 +317,23 @@ def test_SerialConsole_login_except_on_wrong_username_match(serial_console_proxy
...
+    received = 'abcd\nefgh'
+
+    with tempfile.NamedTemporaryFile(delete=False) as tmpfile:
+        console = SerialConsole(
+            port=os.ttyname(pty_pair.main.fd),
+            baud=115200,  # Baud Doesn't really matter as virtual tty,
+            encoding='utf-8', raw_logfile=tmpfile.name)
+        # tmpfile.name
+
+        pty_pair.secondary.write(received)
```
prototypicalpro, line 332: I almost forgot one other modification I made: since the serial port doesn't open until it is first interacted with, I needed to call `require_open` before sending it data.
```suggestion
        console.require_open()
        pty_pair.secondary.write(received)
```

Pixep: Thank you for all these feedback, I just fixed it, and saw the comments after haha. Anyway, yes, I basically have the same thing as you know. I was tired when I wrote this I think.

PR #174, opened 2021-01-08
Static type checking

### LINKED ISSUE #173: Commit to static typing
It's well known that static typing reduces bugs. Python is by it's nature a dynamic language and does well from this.
However, I think we can have the best of both worlds by saying that our code should be statically typed.
What I mean by this is that we use type hinting in all our production code, and have a tool enforce that.

An added and quite significant bonus is that the editor is then able to accurately autocomplete your code for you, and also tell you early if you've done something wrong!

As the project grows and has more contributors, type safety becomes more important.
This is pretty equivalent to [why you should use Typescript for large projects in the Javascript world](https://serokell.io/blog/why-typescript).

**Describe the solution you'd like**
Type hinting on all production code, enforced with: [PyRight](https://github.com/microsoft/pyright)
VS Code's (PyLance)[https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance] extension uses PyRight under the hood.

Let's make this part of our CI pipeline, forcing type safety, we're already most of the way there anyway!
We can start with the gentler option that only says that functions with defined types are type checked, and then later switch to strict typing where all functions MUST have type signatures.

Let's not have our unit tests type checked though. Dynamic programming does have it's advantages!

mjftw:
Closes #173


mjftw:
So the CI _WILL_ fail, as we have loads of type violations.
I think that's okay for now as this PR just adds to tooling to show us those issues.
We should add new issues and work to fix those errors.

mjftw:
> I would say, let's merge it. But we need to make the CI pass (in an "expected fail" style). We will address all of those over time I believe and then enforce it.

Yeah agreed. I don't really like the big red X, but making the CI pass on failure makes the step feel a little pointless.
We need a soft (expected) failure, like is suggested here: https://github.com/actions/toolkit/issues/399

mjftw:
Tried adding `continue-on-error: true` to the CI job, but that made no difference and we still get the big red X. Remove now.

Pixep (approved):
Good to go, rebased

PR #175, opened 2021-01-11
Fix most config related typing issues

Pixep:
Add typed and optional versions of the configuration 'pop' function


pluma/cli/config.py
```diff
@@ -28,26 +29,117 @@ def __init__(self, config: dict = None):
...
+
+    def pop(self, attribute: str, default=None):
         value = self.pop_raw(attribute, default)
         if isinstance(value, dict):
             return Configuration(value)
 
         return value
 
-    def pop_raw(self, attribute, default=None):
+    def pop_optional_str(self, attribute: str, default: Optional[str] = None,
```
mjftw, line 44: Very nice to see type safe versions of these functions to do input validation 👍🏼 

My only suggestion for improvement would be that some of the duplication here could be reduced, perhaps by passing in the type.
We'd maybe end up with something like:

```python
from typing import Optional, TypeVar

T = TypeVar('T')

# (the current pop function could be renamed pop_unsafe, or this one be called pop_safe)
def pop(self, datatype: T, attribute: str, context: Optional[str] = None):
    value = self.pop_optional(attribute, datatype, context=context)
    if value is None:
        self.__raise_error(f'missing required {datatype.__name__} attribute "{attribute}"', context)

    return value

def pop_optional(self, datatype: T, attribute: str, default: Optional[T] = None,
                 context: Optional[str] = None) -> Optional[T]:
    value = self.pop_raw(attribute, default)
    if value is not None and not isinstance(value, datatype):
        self.__raise_error(f'attribute "{attribute}" should be a {datatype.__name__} '
                            f'but got "{value}"', context)
    return value
```

That way the typed calls would like like:

```python
self.pop(bool, my_boolean_name)
# and
self.pop_optional(str, my_str_name, "Hello world!")
```

Could also do this by passing in a predicate function to be used to evaluate whether the value read is valid or not, but this way you don't have the type information to print in the `__raise_error()` call.

E.g. that would look like
```python
def pop(..., predicate: Callable[[any], bool]], ...):
    value = get_the_value()
    if not predicate(value):
        # error handling
    ...
```

And the call site would look like:
```python
self.pop('foo_bar', lambda x: isinstance(x, bool), ...)
```
This would handle special requirements for type checking (E.g. like you're doing with int), but I think it's much uglier.


Anyway, not a hard requirement to abstract this but might make it cleaner.

pluma/cli/config.py
```diff
@@ -28,26 +29,117 @@ def __init__(self, config: dict = None):
...
-    def read_and_keep(self, attribute):
+    def read_and_keep(self, attribute: str):
         return self.config.get(attribute)
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self.config)
 
-    def first(self):
+    def first(self) -> str:
```
mjftw, line 137: For stuff like this where it might raise an exception, we could use type
```
def first(self) -> Union[str, NoReturn]:
    ....
```

This would indicate that the function may not exit with a value at all, but don't know if this is more or less readable?

pluma/core/builder/yoctocbuilder.py
```diff
@@ -63,9 +63,10 @@ def install_yocto_sdk(yocto_sdk: str, install_dir: str = None) -> str:
         return install_dir
 
     @staticmethod
-    def get_yocto_sdk_env_file(install_dir: str) -> Path:
+    def get_yocto_sdk_env_file(install_dir: Path) -> Path:
         '''Return the path to the Yocto SDK environment file in a folder or raise an error.'''
-        install_dir = Path(install_dir)
+        if not isinstance(install_dir, Path):
+            raise Exception('install_dir must be a Path')
```
mjftw, line 69: I'd probably go for a `TypeError` here rather than a bare `Exception`.

mjftw (changes requested):
This is looking really good! Great to see all this lovely type info.

I've added a couple of suggestions for small improvements, mostly just to reduce duplication and allow us to cut a fair bit of code out.

pluma/cli/config.py
Pixep: That's a good point yes. I wanted to go in that direction somehow but kept it simple for now. That should be an easy change to make

pluma/cli/config.py
Pixep: Mhh I have never seen that, is that a common way of expressing that an exception can be raised?

pluma/core/builder/yoctocbuilder.py
Pixep: 👍

pluma/cli/config.py
prototypicalpro: It looks like the docstring of the function is the only way, as the typing system [doesn't support it](https://www.python.org/dev/peps/pep-0484/#exceptions).

pluma/cli/deviceactionregistry.py
```diff
@@ -1,20 +1,22 @@
...
-        def decorator(action_class: DeviceActionBase):
+        def decorator(action_class: Type):
             cls.register_class(action_class=action_class,
                                action_key=action_key)
             return action_class
         return decorator
 
     @classmethod
-    def register_class(cls, action_class: DeviceActionBase, action_key: str):
+    def register_class(cls, action_class: Type, action_key: str):
```
prototypicalpro, line 19: You can specialize `Type` to indicate that `action_class` should be or inherit from `DeviceActionBase`:
```suggestion
    def register_class(cls, action_class: Type[DeviceActionBase], action_key: str):
```

Note that if you want to keep the subclass error check you'll need to replace `action_class.__name__` with `cast(Type, action_class).__name__` to make the type system happy.

pluma/cli/config.py
prototypicalpro: For `pop_optional`, we could do something like this to ensure the return type isn't optional when the default keyword is specified:
```python
    def pop_optional(self, datatype: Type[T], attribute: str, default: D = None,
                 context: Optional[str] = None) -> Union[T, D]:
        value = self.pop_raw(attribute, default)
        if value is None or not isinstance(value, datatype):
            self.__raise_error(f'attribute "{attribute}" should be a {datatype.__name__} '
                                f'but got "{value}"', context)
        return value
```

Note that for pylance I can't get it to resolve the generics D and T correctly, but I'd be curious if it's better with the other tools.

prototypicalpro (changes requested):
One error with pylance I can mention:
https://github.com/Witekio/pluma-automation/blob/fe125d87520acb351bfecd5a4d25fe475ae460e1/pluma/cli/config.py#L239 For some reason pylance doesn't recognize `yaml.parser.ParserError`, but it does recognize `from yaml.parser import ParserError`. Not sure if that's a bug or a python thing.

pluma/cli/config.py
prototypicalpro: Here's an updated version that resolves the types correctly in pylance. I ended up having to create some overload stubs to deal with the type of default so it's a bit uglier, but it should result in less unnecessary None checks when using `pop_optional`.

```python
    # Type if default is None
    @overload
    def pop_optional(self, datatype: Type[T], attribute: str, default: None = None, context: Optional[str] = None) -> Optional[T]: ...
    # Type if default is a non-None value
    @overload
    def pop_optional(self, datatype: Type[T], attribute: str, default: D, context: Optional[str] = None) -> Union[T, D]: ...
    # Implementation
    def pop_optional(self, datatype: Type[T], attribute: str, default: Optional[D] = None,
                 context: Optional[str] = None) -> Union[T, Optional[D]]:
        value = self.pop_raw(attribute, default)
        if not isinstance(value, datatype) and not isinstance(value, type(default)):
            self.__raise_error(f'attribute "{attribute}" should be a {datatype.__name__} '
                                f'but got "{value}"', context)
        return cast(Union[T, D], value)
```

pluma/cli/config.py
Pixep: I agree for a function that never returns anything, or sometimes does not. But any function has the potential to throw if something goes wrong, i.e. all functions need `Union[xxx, NoReturn]`, which would in return provides no added value. So in this situation, I don't see a point in adding it. Different opinions?

pluma/cli/ctestsprovider.py
```diff
@@ -21,28 +23,27 @@ def all_tests(self, key: str, config):
         if not config:
```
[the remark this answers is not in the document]
Pixep: Yes, I think I will go over all providers in a different PR, to not overwhelm it more than it is :)

pluma/cli/deviceactionregistry.py
Pixep: Done

pluma/cli/ctestsprovider.py
```diff
@@ -21,28 +23,27 @@ def all_tests(self, key: str, config):
...
         if not toolchain_file and not env_file:
             raise TestsConfigError(
                 'Missing \'yocto_sdk\' or \'source_environment\' attributes for C tests')
 
-        install_dir = None
-        if toolchain_file:
+        if env_file:
+            env_file = Path(env_file)
+        else:
             install_dir = YoctoCBuilder.install_yocto_sdk(toolchain_file)
```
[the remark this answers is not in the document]
Pixep: Yeah I can't because we need one or the other. In this branch, toolchain_file has to be set (otherwise we go in the one above.

pluma/cli/config.py
Pixep: Thank you both for your help, done :)

Pixep:
Down from 2xx to 157 errors for the type checking

pluma/cli/ctestsprovider.py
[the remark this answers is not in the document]
prototypicalpro: Alright, in that case a cast will get rid of the error:
```suggestion
            install_dir = YoctoCBuilder.install_yocto_sdk(cast(str, toolchain_file))
```

pluma/cli/config.py
prototypicalpro: Something I just realized is that python has a `MutableMapping` base class that implements `popitem` on top of abstract methods: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes. We may want to switch to this implementation since it is definitely the more pythonic way of doing things, but on the other hand if we switch it will require another decent refactor (replace `pop_optional` with `popitem`).

pluma/cli/ctestsprovider.py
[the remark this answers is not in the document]
Pixep: Yes right, done.

pluma/cli/config.py
prototypicalpro: On second pass, it looks like `popitem` doesn't perform any type checking, which is a beneficial aspect of `pop_optional`. Perhaps we could rename `pop_optional` to `popitem_typechecked` (or similar) and have it coexist with `popitem`?

Pixep: AFAIS, popitem is unrelated, and meant to pop a key value pair in a LIFO fashion.

Pixep: Okay, just had another look at it: we may bring them closer, but its going to take more than a few minutes I think, to ponder the implications. I would recommend avoiding pushing to much in that one PR.

mjftw (approved):
Really nice work, this is looking great!

prototypicalpro (approved):
Lgtm!

PR #176, opened 2021-01-11
Bugfix/consolebase no open on match wait

mjftw:
Found this bug when working as a user of Pluma.


tests/generic/core/test_ConsoleBase.py
```diff
@@ -184,4 +184,94 @@ def test_ConsoleBase_send_control_calls_engines_send_control(basic_console):
...
+
+
+def test_ConsoleBase_should_call_open_for_wait_for_match(basic_console):
+    basic_console.close()
+    assert not basic_console.is_open
+
+    basic_console.open = MagicMock(side_effect=basic_console.open)
+
+    basic_console.wait_for_match(['foo'])
+    basic_console.open.assert_called()
```
Pixep, line 197: `assert_called_once` maybe? Otherwise looking good :)

Pixep (approved):

tests/generic/core/test_ConsoleBase.py
mjftw: I thought about that, but then decided that actually I don't think we  really care if `open` got called multiple times as it shouldn't do anything on repeats anyway. Really all these tests care about is that it *was* called.

PR #177, opened 2021-01-11
Expose power controllers reboot_delay in target config

mjftw:


pluma/cli/targetconfig.py
```diff
@@ -190,6 +191,9 @@ def create_power_control(power_config: Configuration, console: ConsoleBase) -> P
                 f'Unsupported power control type "{control_type}". Supported types: {POWER_LIST}')
 
         power_config = power_config.pop(control_type) or Configuration()
+
+        reboot_delay = power_config.pop('reboot_delay')
```
Pixep, line 195: @mjftw Could you add a quick mention in the documentation?

Pixep (changes requested):

pluma/cli/targetconfig.py
mjftw: We've actually not got a section in the docs about the power controllers in the target file at all.
Rather than adding it all now, do you mind if we create a new issue to add this info (for all power controllers)?

pluma/cli/targetconfig.py
```diff
@@ -1,5 +1,6 @@
 import json
 import os
+import re
```
Pixep, line 3: Wait... that's not used, is it?

Pixep (changes requested):

pluma/cli/targetconfig.py
mjftw: Oops. Automatic imports can be really useful, but also not so good sometimes.

mjftw: Fixed.

Pixep (approved):
G2G!

PR #178, opened 2021-01-11
Feature/prepare for pypi push

mjftw:
Once merged, this PR should mean that every push to master also pushes a new python distribution package to Pypi.


setup.py
```diff
@@ -65,11 +65,14 @@ def get_version():
     requires.append('dataclasses')
 
 setuptools.setup(
-    name='pluma-automation',
+    name='pluma',
     version=get_version(),
     author='Witekio',
     author_email='mwebster@witekio.com',
```
mjftw, line 71: FYI on this one, in order to push to Pypi I need a single email address to go here.
Ideally this would also be an email address that users could contact to get support, so currently put mine down but there may be something better. Maybe an email group at Witekio like pluma-dev@witekio.com which we're all on?
That would mean we don't have to update Pypi if I were to get hit by a bus or something.

Pixep: Yes good idea. That would be an IT ticket I believe...?

.github/workflows/bump-version-and-publish-to-pypi.yml
```diff
@@ -0,0 +1,44 @@
+name: Version tag bump and publish to PyPI
```
Pixep, line 1: Is there a way to run that _only_ if the CI passes on master? To prevent/catch issues

Pixep (changes requested):
Just one question regarding publishing only if all succeeds

.github/workflows/bump-version-and-publish-to-pypi.yml
mjftw: That's a very good idea.

The difficulty here is that for the push to PyPi to happen, we need to have bumped the version tag to something proper else it will get rejected. As far as I could see (from a few mins of googling) there's no way to make one Github Action depend on another, and the expected way is to do it all with different jobs in the same action. It's for this reason that I've put the publish CI in the same Action as the tagging job.
If we can't make one Action/Workflow depend on another then the only way to do your suggestion would be to do the tag and push in the same Action/Workflow as all of the other CI checks we want to go before, which is a bit messy and quite a big change.
Also, if commits are merged to master we do _need_ there to be a new version tag, else it's impossible to know where one feature changeset ends and another begins (because we are doing rebase merges).

The counter argument to this need would be, if the CI passed on the pull request to master, but failed when merged to master, does that not mean that the issue is with the CI and merging strategy rather than the code under test?

setup.py
mjftw: IT ticket raised

.github/workflows/bump-version-and-publish-to-pypi.yml
Pixep: > The counter argument to this need would be, if the CI passed on the pull request to master, but failed when merged to master, does that not mean that the issue is with the CI and merging strategy rather than the code under test?

If we only do fast forward yes. Maybe that is the easy answer: no rebase from GitHub, only fast forwards :)

Pixep (approved):
Good to go

.github/workflows/bump-version-and-publish-to-pypi.yml
mjftw: Good idea.
Since we'll be relying on people don't the right kind of merges, I'll change the settings to enforce that

PR #182, opened 2021-01-14
CI: Fetch git tags before attempting pypi push

mjftw:
The CI is failing to push to Pypi because when it get to the point where it generates the Pluma wheel distro to be pushed it reads the Pluma version as some random commit hash.
It looks like this is because github doesn't do a full clone of the repository and does not pull the git tags.
The git tags are used by Pluma's `setup.py` script to determine what the current version is using `git describe`. No tags means no valid version tag to use as the version, which means we get a random commit hash as the package version.
A random commit hash is obviously not a valid python package version so Pypi says "No.".

The solution is likely to be just fetching the git tags before building the Pluma distro, so that the correct version can be found.
I've done some testing in another branch and I think this will work, but can't really test for real until we merge this PR into master, kicking off the process for real.


prototypicalpro:
In case that doesn't work, another option is modifying your `@actions/checkout` action to pull the tags:
```yml
- uses: actions/checkout@master
  with:
    fetch-depth: 0
```
or creating another workflow file that runs on tag push:
```yml
on:
  push:
    tags:
    - 'v*'
```

mjftw:
> In case that doesn't work, another option is modifying your `@actions/checkout` action to pull the tags:
> 
> ```yaml
> - uses: actions/checkout@master
>   with:
>     fetch-depth: 0
> ```

That's a much nicer solution!
Swapping to this suggestion 👍🏼

Pixep (approved):
Thank you for fixing it :)

PR #187, opened 2021-01-19
Console: fix type hints issues in console related files

Pixep:
Removes another 40-ish type hints warnings


pluma/core/baseclasses/consolebase.py
```diff
@@ -58,17 +58,25 @@ def read_all(self, preserve_read_buffer: bool = False) -> str:
...
+    def wait_for_match(self, match: List[str], timeout=None) -> Optional[str]:
         '''Wait a maximum duration of 'timeout' for a matching regex, and returns matched text'''
         self.require_open()
         match_result = self.engine.wait_for_match(match=match, timeout=timeout)
         return match_result.text_matched
 
-    def wait_for_bytes(self, timeout: int = None, sleep_time: int = None,
+    def wait_for_bytes(self, timeout: Union[int, float] = None,
+                       sleep_time: Union[int, float] = None,
                        start_bytes: int = None) -> bool:
```
prototypicalpro, line 69: The `int` type is included in the `float` type:
```suggestion
    def wait_for_bytes(self, timeout: Optional[float] = None,
                       sleep_time: Optional[float] = None,
                       start_bytes: Optional[int] = None) -> bool:
```

pluma/core/baseclasses/consolebase.py
```diff
@@ -126,8 +134,9 @@ def wait_for_quiet(self, quiet: float = None, sleep_time: float = None,
         # Timeout
         return False
 
-    def send_and_read(self, cmd: str, timeout: int = None,
-                      sleep_time: int = None, quiet_time: int = None,
+    def send_and_read(self, cmd: str, timeout: Union[int, float] = None,
+                      sleep_time: Union[int, float] = None,
+                      quiet_time: Union[int, float] = None,
```
prototypicalpro, line 139:
```suggestion
    def send_and_read(self, cmd: str, timeout: Optional[float] = None,
                      sleep_time: Optional[float] = None,
                      quiet_time: Optional[float] = None,
```

prototypicalpro (changes requested):
Some minor typing fixes. Good to see the type error count come down!

pluma/core/baseclasses/consolebase.py
Pixep: Haha of course...😌 thanks

pluma/core/baseclasses/consolebase.py
Pixep: Done

pluma/core/baseclasses/consolebase.py
```diff
@@ -58,17 +58,25 @@ def read_all(self, preserve_read_buffer: bool = False) -> str:
...
 
-    def wait_for_bytes(self, timeout: int = None, sleep_time: int = None,
+    def wait_for_bytes(self, timeout: Optional[float] = None,
+                       sleep_time: Optional[float] = None,
                        start_bytes: int = None) -> bool:
         '''Wait for data to be received on the console.'''
-        timeout = timeout or 10.0
-        sleep_time = sleep_time or 0.1
+        if timeout is not None:
+            timeout = timeout*1.0
```
mjftw, line 72: Is this to force float conversion?

If so I'm guessing it's here to keep the type checker happy?
I can see why you'd do it, but it does obfuscate the code a little.

pluma/core/baseclasses/pexpectengine.py
```diff
@@ -98,9 +101,11 @@ def wait_for_match(self, match: List[str],
             matched_regex = None
             matched_text = None
 
-        return MatchResult(regex_matched=matched_regex,
-                           text_matched=self.decode(matched_text),
-                           text_received=self.decode(received))
+        text_matched = self.decode(matched_text) if matched_text else None
```
mjftw, line 104: Good spot! Shouldn't be decoding this if it's None.

tests/generic/core/test_PexpectEngine.py
```diff
@@ -151,6 +151,15 @@ def test_PexpectEngine_send_control_sends_correct_bytes(pty_pair_raw):
 
     assert pty_pair_raw.main.read(timeout=0.5) == b'^C'
 
+
+def test_PexpectEngine_error_on_control_longer_than_one_char(pty_pair_raw):
```
mjftw, line 155: Aaah yeah good test to add, didn't think to add that one

mjftw (commented):
This is looking like a big improvement, thanks!
Do you think we'd be able to find away around that implicit float conversion line? (`x = x * 1.0`)

pluma/core/baseclasses/consolebase.py
mjftw: Just making it more explicit what we're doing might be enough if there's no way to keep the type checker happy otherwise.
```python
if timeout is not None:
    timeout = float(timeout)
```

pluma/core/baseclasses/pexpectengine.py
Pixep: It was previously accepting `None`, and working fine. But in the end the type check was being a bit annoyed by it, so I changed the way this works :)

pluma/core/baseclasses/consolebase.py
Pixep: Yes that's now better, I was able to revert to the original form, now that the input type is only `Optional[float]`. Before this wasn't possible. Fixed.

mjftw (approved):
Looking great, merging 👍🏼

PR #188, opened 2021-01-20
Fix type hints for TestController and TestRunner

Pixep:


pluma/test/testrunner.py
```diff
@@ -117,7 +117,7 @@ def add_test(self, test: TestBase):
         '''Add a test to the test list. Handles tests with same name by appending a number'''
         # Check if user accidentally passed in a class inheriting
         # TestBase, instead of an instance of that class.
-        if inspect.isclass(test) and (test is TestBase or issubclass(test, TestBase)):
+        if inspect.isclass(test) and issubclass(test.__class__, TestBase):
```
mjftw, line 120: Much better

mjftw (approved):
Another easy approval!

pluma/test/testrunner.py
Pixep: Actually this is wrong, made the typecheck happy, but I completely missed the point here haha. I saw that in the last PR I created. This is to test if `test` is actually a TestBase class (not instance), which makes the typecheck super not happy. Because we have the typecheck though, I think we can safely remove it right?

pluma/test/testrunner.py
```diff
@@ -104,7 +103,7 @@ def _init_test_data(self, test: TestBase):
         }
 
     @property
-    def tests(self):
+    def tests(self) -> List[TestBase]:
```
prototypicalpro, line 106: Since the `tests` argument in `__init__` accepts an iterable of TestBase, should the return type for this be `Iterable[TestBase]` as well? Or am I missing a conversion?

Pixep: We are storing it as a List, but accept an Iterable to populate the list

prototypicalpro (approved):

PR #190, opened 2021-01-21
Plan, Session, and TestGroup

Pixep:
Add a few new classes to support future changes:

* TestGroup now responsible for holding/adding/removing tests (removed from TestRunner)
* GroupedTest inherits TestBase and provides a generic way to run tests transparently (regardless of if it is a group or not)
* Session inherits GroupedTest, restarts and power off the device
* Plan inherits GroupedTest, and is just that.


pluma/test/testgroup.py
```diff
@@ -0,0 +1,112 @@
...
+            self._tests.remove(test)
+
+    def get_test_by_name(self, test_name: str) -> Optional[TestBase]:
+        tests_with_name = [t for t in self.tests if str(t) == test_name]
+        if len(tests_with_name) > 1:
+            raise TestingException(f'Found multiple tests with name {test_name}')
+
+        return None if not tests_with_name else tests_with_name[0]
+
+    def __give_unique_name(self, test: TestBase) -> TestBase:
```
Pixep, line 70: This was copied from TestRunner, but I'm not too happy about it. First because of the implementation :), and second because it is only going to be unique within this TestGroup. If we are looking for unique identifiers, then we might need a central test/task ID registry, instead of relying purely on the name. Not urgent to decide, and probably a different ticket, but will need to change I think.

pluma/test/testgroup.py
```diff
@@ -0,0 +1,112 @@
...
+                                       f'tests with name [{original_name}] reached!')
+
+        return test
+
+
+class TestGroup(TestGroupBase, TestBase):
+    '''A set of tests which can be handled as a test'''
+
+    def __init__(self, name: str = None, tests: Union[TestBase, List[TestBase]] = None):
+        super(TestGroup, self).__init__()
```
mjftw, line 97: I'm never sure about the use of `super()` with multiple inheritance.
Whose `__init__()` gets called here; TestGroupBase, or TestBase? Both?

I usually tend to be explicit and do something like:

```python
TestBase.__init__(self)
TestGroupBase.__init__(self)
```

Although I know this goes against the generally accepted good style.

tests/generic/cli/test_cli.py
```diff
@@ -179,7 +180,9 @@ def test_cli_should_create_results_file_with_correct_testrunner_data(pluma_cli,
         "example_plugin.maths.Maths": {
             "tasks": {
                 "ran": [
-                    "test_body"
+                    "setup",
```
mjftw, line 183: Would we only have "setup" and "teardown" included in the output here if the functions were actually defined in the test that ran? As in, if the test implementation doesn't have a setup or teardown it wouldn't get listed here.

pluma/test/session.py
```diff
@@ -0,0 +1,8 @@
+from pluma.test import TestGroup, TestList
+
+
+class Session(TestGroup):
+    '''A group of tests organized as a session'''
```
mjftw, line 5: What's the difference between a `Session` and a `TestGroup`?
Would be good to make it really clear in the docstring here.

pluma/test/testgroup.py
```diff
@@ -0,0 +1,112 @@
...
+
+from pluma.test import TestBase, TestingException
+from pluma.core.baseclasses import Logger
+
+TestList = List[TestBase]
+
+log = Logger()
+
+
+class TestGroupBase:
```
mjftw, line 13: Abstract (inherits ABC)?

pluma/test/testgroup.py
Pixep: I see, apparently the way I wrote it is the generally recommended way, it is going to resolve the order in which to call constructors (direct parents first, then others).

pluma/test/session.py
Pixep: Right now none :). Session will have more in the future though, probably rebooting the device between each session.

pluma/test/testgroup.py
Pixep: Yeah exactly. But I think understanding better the requirement for unique names/id would be good too first. Do you remember what the core goal was behind having unique names?

pluma/test/testgroup.py
Pixep: Nevermind, I got rid of this this morning, didn't like it too much anyway. No more multiple inheritance, better :)

pluma/test/testgroup.py
```diff
@@ -0,0 +1,112 @@
...
+    '''A named set of tests'''
+
+    def __init__(self, name: str = None, tests: Union[TestBase, List[TestBase]] = None):
```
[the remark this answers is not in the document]
Pixep: I kept this one concrete, but it is no more inherited

tests/generic/cli/test_cli.py
Pixep: Not anymore, because TestBase is now an ABC, with setup and teardown as abstractmethods. This was a side effect of the TestRunner using `hasattr`. I think having them in TestBase rather than arbitrary functions makes more sense (especially for these two), knowing they always exist. Ideally we wouldn't run them, but I don't know if its a big deal.

pluma/test/testgroup.py
Pixep: No more Base class

pluma/test/testgroup.py
prototypicalpro: I'm not sure if this is the core goal, but the PlanRunner XRay integration associates each test in the configuration with a JIRA test ticket using the name of the test, which means that the names must be unique and should probably be deterministic (to prevent unnecessary ticket creation). One idea (which I am probably going to do for the xray integration) is using the config line and test name: `ShellTest@pluma.yml:23`. We could also require the user to supply their own name, but that's not the most friendly approach.

Pixep: Thanks, good input. The name of the file won't be enough however, because some test are definitely reused multiple times with the same or different parameters :). But I agree this should be deterministic.

Pixep: Addressed by #196

tests/generic/cli/test_cli.py
Pixep: Edit: Tried running only if not "empty", I find is more confusing than simply always running those actually.

mjftw: Yeah fair enough, I don't think it's a big deal.
The issue here is probably more how it's portrayed on the reporting, and we're changing that anyway.

pluma/test/testgroup.py
```diff
@@ -0,0 +1,81 @@
...
+            self.add_test(test)
+
+    def add_test(self, test: TestBase):
+        '''Add a test to the test list. Handles tests with same name by appending a number'''
+        if not isinstance(test, TestBase):
+            raise TypeError('The test must be a TestBase instance, '
+                            f'but got {test} instead.')
+
+        if self.get_test_by_name(str(test)) is not None:
+            raise RuntimeError(f'Found duplicate test name {str(test)}!'
```
mjftw, line 50: Very nice addition!

pluma/test/testrunner.py
```diff
@@ -60,16 +58,15 @@ def run(self) -> bool:
...
-        self.log("Running tests: {}".format(
-            list(map(str, self.tests))), level=LogLevel.DEBUG)
+        self.log(f'Running tests: {list(map(str, self.tests))}', level=LogLevel.DEBUG)
 
         try:
             # Defer the actual test running to classes that inherit this base
             self._run(self.tests)
-
-        # Prevent exceptions from leaving test runner
         except Exception:
```
mjftw, line 66: Seeing as we're modifying this bit of the code anyway, we could log the type and message of the exception maybe

pluma/test/plan.py
```diff
@@ -0,0 +1,13 @@
+from pluma.core import Board
+from pluma.test import GroupedTest, TestList
+
+
+class Plan(GroupedTest):
```
mjftw, line 5: Is a `Plan` basically no different from a `GroupedTest`, but with different expected usage so therefore a different class for clarity and  to allow divergence?

pluma/test/testrunner.py
```diff
@@ -156,15 +115,15 @@ def _run_tasks(self, tests: Union[TestBase, Iterable[TestBase]],
...
         for test, task in ((test, task)
                            for task in task_names
-                           for test in tests
-                           if hasattr(test, task)):
-            self._run_task(task, test)
+                           for test in tests):
+            self._run_task(test, task)
 
-    def _run_task(self, task_name, test: TestBase):
+    def _run_task(self, test: TestBase, task_name: str):
```
mjftw, line 121: We can leave for now, but since you've simplified the test stuff underneath I think this method and the class in general is ripe for refactoring to something simpler.

tests/generic/test/test_TestGroup.py
```diff
@@ -0,0 +1,111 @@
...
+            pass
+
+    class MyTest2(TestBase):
+        def test_body(self):
+            pass
+
+    group.add_test(MyTest1(mock_board))
+    group.add_test(MyTest2(mock_board))
+
+    assert len(group) == 2
```
mjftw, line 76: Ooooo very nice!

mjftw (approved):
Really nice code and extensive tests, a great PR 👍🏼 
image

prototypicalpro (approved):
I can't say that I understand all of what's happening (not yet anyway), but the parts that I do understand look good 👍. 
I have one comment that is probably best left to a different PR, and that is I notice that for keyword arguments the `Optional` is often omitted:
```python
def myfunc(arg: str = None)
```
Apparently this _was_ valid syntax but has since been [phased out](https://www.python.org/dev/peps/pep-0484/#union-types), and as a result mypy will correctly add `Optional` to the parameter type but pyright will not. Because of this, in the future we will want to find all these parameters and update their types so pyright can check them correctly. Since this problem is outside the scope of this PR, I will go ahead and open an issue for the backlog.

pluma/test/plan.py
Pixep: Yep exactly

pluma/test/testrunner.py
Pixep: Agree, this will be in a next PR

pluma/test/testrunner.py
Pixep: Yes, good idea.

Pixep:
> Apparently this _was_ valid syntax but has since been [phased out](https://www.python.org/dev/peps/pep-0484/#union-types), and as a result mypy will correctly add `Optional` to the parameter type but pyright will not. Because of this, in the future we will want to find all these parameters and update their types so pyright can check them correctly. Since this problem is outside the scope of this PR, I will go ahead and open an issue for the backlog.

I was hoping we would not need too, given how verbose the signatures will become, but it seems that we will have to. Feel free to create a ticket with the details yes, please :). Thanks.

PR #192, opened 2021-01-22
Fix all type hints (kind of)

Pixep:
Clear all remaining errors except a couple, and a few files which we have to choose to drop, or refactor:

*  hub.py, modem.py, pdu.py, email.py, git.py

CI passing for type check.

@mjftw: Do you have an opinion on these?


pluma/utils/helpers.py
```diff
@@ -22,32 +23,10 @@ def run_host_cmd(command, stdin=None, *args, **kwargs):
...
         kwargs_final.update(kwargs)
 
-    child_proc = subprocess.Popen(
-        command.split(), *args, **kwargs_final)
-    return (child_proc.communicate(
-        input=None if not stdin else bytes(stdin, 'utf-8')
-    )[0].decode('utf-8'), child_proc.returncode)
-
-
-def format_json_tinydb(db_file, db_output_file):
```
Pixep, line 32: dropped, as unused

pluma/core/baseclasses/powerbase.py
```diff
@@ -1,28 +1,36 @@
-from .hardwarebase import HardwareBase
 import time
 from abc import ABCMeta, abstractmethod
+from .hardwarebase import HardwareBase
 from functools import wraps
 
 
+def power_on(f):
```
Pixep, line 7: These were messed up, actually `abstractmethod` which were also decorators, both with different signatures.

pluma/test/testrunner.py
```diff
@@ -36,7 +36,8 @@ def __init__(self, board: Board = None, tests: Union[TestBase, Iterable[TestBase
         self._tests = []
         if isinstance(tests, TestBase):
             tests = [tests]
-        self.tests = tests
+
+        self.tests = tests  # type: ignore
```
mjftw, line 40: ??

pluma/utils/email.py
```diff
@@ -1,3 +1,6 @@
+# type: ignore
+# TODO: Drop file, or fix type check
```
mjftw, line 2: ??

pluma/utils/git.py
```diff
@@ -232,16 +232,20 @@ def compile_version_list(srcdir, version_filters):
         return None
 
     if filtered_versions_lists:
-        filtered_versions = set(filtered_versions_lists[0])
+        # TODO: Are we sure [0] is never None?
```
mjftw, line 235: Yeah, we can be pretty confident of that at this point since the previous block ensures that.

```python
    if None in filtered_versions_lists:
        return None
        return None
```

pluma/utils/git.py
```diff
@@ -232,16 +232,20 @@ def compile_version_list(srcdir, version_filters):
...
+        filtered_versions = set(filtered_versions_lists[0])  # type: ignore
```
mjftw, line 236: I don't really like having all these `#type: ignore` bits in here (or anywhere really).
Feels like if we can't fix all the type errors right now then we should just let these lines fail and fix them later, rather than artificially making it look like there are no errors, otherwise it's unlikely we'll ever remember these are here and they'll never get fixed.
My other reason for disliking them is that they're at a completely different level of abstraction/ purpose than all of the code around them, which hurts readability and is bad for single responsibility principle. I'm not a fan of all inline external tool pragmas (which is pretty much what this is) for this reason.

I can see why you'd stick them in to try to get the type check passing, but maybe better to let it fail for now.

pluma/utils/helpers.py
mjftw: I think this was used by a user of the Pluma library (a different project) rather than the library itself.
You could make the argument that it belongs in the other project should be implementing this though if it doesn't really have any baring on Pluma.

mjftw: Oh yeah I remember, this was used extensively for persisting results in another project and was very useful.
We can drop it here though.

tests/generic/core/test_PowerMulti.py
```diff
@@ -11,12 +11,12 @@ def __init__(self, mocks_on_list=None, mocks_off_list=None):
         self.on_called = 0
         self.off_called = 0
 
-    @PowerBase.on
+    @power_on
```
mjftw, line 14: See notes in `powerbase.py`

mjftw (changes requested):
This is looking so much better, and I can see a load of genuine possible bugs that were fixed while doing this :)
Nice one putting the effort in for this. I know fixing compiler errors can be a bit dull.

Some comments below.

mjftw:
> Clear all remaining errors except a couple, and a few files which we have to choose to drop, or refactor:
> 
>     * hub.py, modem.py, pdu.py, email.py
> 
> 
> CI passing for type check.
> 
> @mjftw: Do you have an opinion on these?

What's it complaining about?

The reason for these files is to allow the import paths to be shorter by making the deeper functions/classes/... accessible at the higher levels.

Pixep:
> What's it complaining about?
> 
> The reason for these files is to allow the import paths to be shorter by making the deeper functions/classes/... accessible at the higher levels.

> This is looking so much better, and I can see a load of genuine possible bugs that were fixed while doing this :)
> Nice one putting the effort in for this. I know fixing compiler errors can be a bit dull.
> 
> Some comments below.

To explain my reasoning here: I am pretty much convinced at this point that we **need** a OK/NOK from the CI **now**, as this proved super useful for me when implementing the next TestGroup classes, to catch errors, and we need it given the size of the code base, as we reachitecture parts and bits of it. But this **only** works if the output is clean, if there are no errors reported. Otherwise, we won't bother checking the output, its too painful, and we won't know what was an error before, and what's new, because type errors can show up in any files anyway. That's why I want to reach "0 errors" ASAP, even if this means hiding some, and creating tickets to fix them.

Regarding these 4 files, these are older files, that have tons of typing errors. I probably fixed ~150 alone, but can't really spend more time addressing these, because I know these classes less, and I don't know if they are/will be useful after our merge.

So the best way to have the typecheck available for you/Noah/myself, is to ignore those ~50 remaining errors, to let us focus on what matters, and create tasks to either fix these issues (if the classes will be used) or drop the related code, if it won't be used.

I added the `make validate` as well, which ensures pyright and unit tests pass, which catches so many issues, but again, the output needs to be clean for that.

pluma/utils/helpers.py
Pixep: Ok, its still in Git if needed too :)

pluma/core/baseclasses/powerbase.py
Pixep: Yeah that's actually what I wanted to do, but I thought you wouldn't like it, not sure its the pythonic way of doing this :). But well, its in now!

pluma/core/baseclasses/powerbase.py
```diff
@@ -1,28 +1,36 @@
...
 
     @abstractmethod
-    def on(f):
-        @wraps(f)
-        def wrap(self, *args, **kwargs):
-            self.log(f'{str(self)}: Power on')
-            f(self, *args, **kwargs)
-        return wrap
+    def on(self):
+        '''Turn on the power controll'''
```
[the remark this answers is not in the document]
Pixep: Done

pluma/core/baseclasses/relaybase.py
```diff
@@ -1,15 +1,20 @@
+from abc import ABC, abstractmethod
+
 from .hardwarebase import HardwareBase
-from abc import ABCMeta, abstractmethod
 from functools import wraps
 
 
-class RelayBase(HardwareBase, metaclass=ABCMeta):
+def relay_toggle(f):
```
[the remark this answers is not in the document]
Pixep: Done

pluma/core/baseclasses/powerbase.py
```diff
@@ -1,28 +1,36 @@
...
 
     @abstractmethod
-    def off(f):
-        @wraps(f)
-        def wrap(self, *args, **kwargs):
-            self.log(f'{str(self)}: Power off')
-            f(self, *args, **kwargs)
-        return wrap
+    def off(self):
+        '''Turn off the power controll'''
```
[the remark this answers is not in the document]
Pixep: Done

pluma/core/baseclasses/powerbase.py
mjftw: Oh okay cool :D

mjftw:
> > What's it complaining about?
> > The reason for these files is to allow the import paths to be shorter by making the deeper functions/classes/... accessible at the higher levels.
> 
> > This is looking so much better, and I can see a load of genuine possible bugs that were fixed while doing this :)
> > Nice one putting the effort in for this. I know fixing compiler errors can be a bit dull.
> > Some comments below.
> 
> To explain my reasoning here: I am pretty much convinced at this point that we **need** a OK/NOK from the CI **now**, as this proved super useful for me when implementing the next TestGroup classes, to catch errors, and we need it given the size of the code base, as we reachitecture parts and bits of it. But this **only** works if the output is clean, if there are no errors reported. Otherwise, we won't bother checking the output, its too painful, and we won't know what was an error before, and what's new, because type errors can show up in any files anyway. That's why I want to reach "0 errors" ASAP, even if this means hiding some, and creating tickets to fix them.
> 
> Regarding these 4 files, these are older files, that have tons of typing errors. I probably fixed ~150 alone, but can't really spend more time addressing these, because I know these classes less, and I don't know if they are/will be useful after our merge.
> 
> So the best way to have the typecheck available for you/Noah/myself, is to ignore those ~50 remaining errors, to let us focus on what matters, and create tasks to either fix these issues (if the classes will be used) or drop the related code, if it won't be used.
> 
> I added the `make validate` (#194) as well, which ensures pyright and unit tests pass, which catches so many issues, but again, the output needs to be clean for that.

Oh yeah okay fair enough, that makes sense.

In that case, since we've already got an issue up for this let's go ahead and merge 👍🏼 

You're right, having type check on new PRs is way more important than getting to 100% on old stuff right now.

mjftw (approved):
Great to see so many fixed bugs in with these type checks!

PR #194, opened 2021-01-24
Makefile: add validate and scope for test cmd

Pixep:
Add a `make validate` command to run unit tests and type check


Makefile
```diff
@@ -28,14 +28,16 @@ install-devtools::  ## Install Pluma additional development tools
 		@./install_devtools.sh
 
 test:: ## Run the Pluma Automation tests. Tests in tests/rpi are ignored if not on Raspberry Pi
-		@./tests/scripts/run_tests.sh
+		@./tests/scripts/run_tests.sh $(scope)
```
mjftw, line 31: Really nice little addition that I don't know why I didn't add months ago.
I've been using the `run_tests` script directly a lot of the time so I can specify the scope :D

mjftw (approved):
Great idea!

prototypicalpro (approved):
Looks good! My only comment is that it looks like we attempt to install pyright in `install_devtools.sh` using `sudo npm install -g pyright`, which is a bit dangerous as npm is not designed to run as root (see [resolving EACCESS errors](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally)). It seems like common practice to run npm as root anyway however (especially in containers), so I think either way would work.

Pixep:
> Looks good! My only comment is that it looks like we attempt to install pyright in `install_devtools.sh` using `sudo npm install -g pyright`, which is a bit dangerous as npm is not designed to run as root (see [resolving EACCESS errors](https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally)). It seems like common practice to run npm as root anyway however (especially in containers), so I think either way would work.

Mhh, might be worth going to a different ticket, as this was added in a different PR. This is merely documenting it.

Issue #123, opened 2020-10-05
Expose interactive control of boards to the CLI


mjftw:
**Is your feature request related to a problem? Please describe.**
When Pluma was purely a Python library, a common use case was to remote control of a board's power, console etc. via an interactive python shell.
To this end, the Console classes have an `interact()` method, which drops the user into an interactive shell on the board's console.

**Describe the solution you'd like**
It would be good to expose this (and other functions like power control) to an interactive Pluma shell.

PR #196, opened 2021-01-26
feat: add unique name generation to TestBase

prototypicalpro:
This PR reworks `TestBase#_test_name` to ensure that every test name is unique, and removes the equivalent implementations in `ShellTest` and `DeviceActionBase`. Because this generated test name is no longer consistent across runs, some tests were also reworked to only match the output values and ignore the names.


pluma/test/shelltest.py
```diff
@@ -9,14 +9,11 @@
 
 
 class ShellTest(TestBase):
-    '''Execute script within the target (or host) shell'''
-    shell_test_index = 0
-
-    def __init__(self, board: Board, script: Union[str, List[str]], name: str = None,
+    def __init__(self, board: Board, script: str, name: str = None,
```
Pixep, line 12: Why dropping the type hint here? `script: Union[str, List[str]]`  Typo?

Pixep, line 12: Why dropping the type hint here? `script: Union[str, List[str]]` Typo?

pluma/test/testbase.py
```diff
@@ -15,9 +16,10 @@ def __init__(self, board: Board = None, test_name_suffix: str = None):
         else:
             self._test_name = self.__class__.__name__
 
-        if test_name_suffix:
-            assert isinstance(test_name_suffix, str)
-            self._test_name += f'_{test_name_suffix}'
+        TestBase.test_count += 1
+        if test_name:
+            self._test_name += f'[{test_name}]'
+        self._test_name += f'[{TestBase.test_count}]'
```
Pixep, line 22: Maybe instead `Module.MyTest[abcd]#123`?

pluma/test/testrunner.py
```diff
@@ -4,7 +4,7 @@
 import re
 from abc import ABC, abstractmethod
 from copy import copy
-from typing import Iterable, Optional, List, Union
+from typing import Iterable, List, Union, Optional
```
Pixep, line 7: What?! :D

tests/generic/test/test_TestRunner.py
```diff
@@ -328,8 +329,8 @@ def teardown(self):
         def random_func_not_hook(self):
             pass
 
-    expected_data = {
-        'test_TestRunner.MyTest': {
```
Pixep, line 332: Mhh I see, if you need to remove it everywhere, do you still have a test somewhere that asserts that the test name in the data is valid?

tests/generic/utils.py
```diff
@@ -47,3 +48,47 @@ def nonblocking(f, *args, **kwargs):
     pool = ThreadPool(processes=1)
     async_result = pool.apply_async(f, args=args, kwds=kwargs)
     return async_result
+
+
+class PlumaOutputMatcher:
```
Pixep, line 53: Nice :)

Pixep (changes requested):
Minor comments, looking good

pluma/test/testrunner.py
prototypicalpro: A controversial decision for sure, but one I'm willing to make.

pluma/test/shelltest.py
prototypicalpro: Good catch, that must've gotten messed up in the rebase.

tests/generic/test/test_TestRunner.py
prototypicalpro: I added an name validation step to `PlumaOutputMatcher`, does that address the issue?

Pixep: Yes thanks. Wasn't necessary for all of these, it was very redundant (but necessary). But that works :)

pluma/test/shelltest.py
```diff
@@ -9,14 +9,11 @@
 
 
 class ShellTest(TestBase):
-    '''Execute script within the target (or host) shell'''
```
Pixep, line 12: Don't drop the class docstring :)

pluma/test/testbase.py
```diff
@@ -15,9 +16,10 @@ def __init__(self, board: Board = None, test_name_suffix: str = None):
         else:
             self._test_name = self.__class__.__name__
 
-        if test_name_suffix:
-            assert isinstance(test_name_suffix, str)
-            self._test_name += f'_{test_name_suffix}'
+        TestBase.test_count += 1
+        if test_name:
+            self._test_name += f'[{test_name}]'
+        self._test_name += f'#{TestBase.test_count}'
```
Pixep, line 22: Could you cover this with a quick UT? Then we're good

Pixep (changes requested):
One UT, and restoring one docstring, and good to go!

pluma/test/testbase.py
prototypicalpro: This look good? https://github.com/Witekio/pluma-automation/blob/547cdaffeb175a2327a36de2b506b04d79fc7d7f/tests/generic/test/test_TestBase.py#L4-L21

prototypicalpro: Forgot `_test_name` should be accessed using `str`. How about this: https://github.com/Witekio/pluma-automation/blob/db5dee056b47becad7d107f506fa1492d4cfe0cb/tests/generic/test/test_TestBase.py#L4-L23

Pixep: Yep :)

Pixep (approved):
Thanks, all good!

PR #200, opened 2021-02-02
Prevent Networking python base test from being instantiated

Pixep:


mjftw (commented):
Approving anyway, but yeah right to just go ahead with the merge, very inoffensive PR

PR #201, opened 2021-02-03
Feature/rest api

mjftw:
Work in progress, but some of the way to having a basic REST API for Pluma.
I'll use this PR to give some early exposure to what I'm working on as I go.

This would make integration with CI tools *much* easier, as the CI could just make a POST request to Pluma running on a given machine in order to run a test job.

Step 1 was to abstract the CLI to move all of the filesystem and environment access to the top level. This will allow the `Pluma.execute_tests` etc, to be used from the REST API too.

Step 2 was to add the server framework

Step 3 is to add all the routes for the server, actually building the API.

I'm using FastAPI and Uvicorn because they're awesome and give you SO MUCH stuff for free.

https://fastapi.tiangolo.com/
https://www.uvicorn.org/

Interested to hear your thoughts on this.


Pixep:
Sweet! Then we just need to find a way to make it available publicly online from a local location (I suppose there may be some solutions for that?)

mjftw:
Moved some of this work over to https://github.com/Witekio/pluma-automation/pull/211 in order to get it merged sooner.

PR #202, opened 2021-02-03
Fix IperfBandwidth regex - assumed MBps with decimal point

mjftw:
Issue found by a user when running the `IperfBandwidth` test.

```shell
Task failed: Failed to match iperf bandwidth regex for output:
------------------------------------------------------------
Server listening on TCP port 5001
TCP window size:  128 KByte (default)
------------------------------------------------------------
[  4] local 192.168.1.229 port 5001 connected with 192.168.1.127 port 33216
[ ID] Interval       Transfer     Bandwidth
[  4]  0.0-30.3 sec   275 MBytes  76.0 Mbits/sec
```

This should PR fixes the regex so that it no longer assumes a decimal point in the output. 
Not tested on real hardware yet, but https://regex101.com/r/uAy3PR/1/ implies this is correct.


Pixep:
Oh, nice. Won't that extract the transfer size though in this case?

mjftw:
> Oh, nice. Won't that extract the transfer size though in this case?
Ummmm yes I suppose you are right, do you think regex changes are required?

mjftw:
> > Oh, nice. Won't that extract the transfer size though in this case?
> 
> Ummmm yes I suppose you are right, do you think regex changes are required?

https://regex101.com/r/y4z335/1

Pixep:
Yes, just tested, works for me too, thanks!

Pixep (approved):

PR #203, opened 2021-02-03
Expose SoftPower reboot_cmd in target config

mjftw:
Having an `on_command` seems a little counter intuitive. If the board is off, how do you send a command to it?
A reboot command makes much more sense, so I've exposed that to the target config to make it usable.


Pixep (approved):
Looks good, thank you! Still need to document all that

PR #204, opened 2021-02-03
Fix/soft power improvements

Pixep:
Various small improvements, mainly fixing the test used for the default reboot cmd to be used.


mjftw (approved):
Nice one 👍

PR #206, opened 2021-02-03
CI: automatically publish docker image from master

Pixep:


mjftw (commented):
Excellent, I've wanted this for a while!!

Will this push the container with a version matching that of the Pluma version?

Pixep:
> Excellent, I've wanted this for a while!!
> 
> Will this push the container with a version matching that of the Pluma version?

Only with the `:latest` tag for now. Because tagging is done in a different action, might need some changes too. But at least allows to have the latest/master available

mjftw:
> > Excellent, I've wanted this for a while!!
> > Will this push the container with a version matching that of the Pluma version?
> 
> Only with the `:latest` tag for now. Because tagging is done in a different action, might need some changes too. But at least allows to have the latest/master available

Okay. Yeah let's get this merged, but would be good to get tagging in there too ASAP so we get the history up there as knowing which version of Pluma your container is for is quite important with the frequency we're making changes.

mjftw (approved):

PR #207, opened 2021-02-04
install.sh: ensure setuptools is available before installing

Pixep:


Pixep:
Regarding a problem Nick had on a fresh Ubuntu, apparently setuptools was not provided.

mjftw (approved):
Good spot!

nelsmore:
I know this is closed, but I also had an issue with wheel.  Should this be installed too?  On Ubuntu 18.04 I was unable to properly install without first manually installing wheel.

Pixep:
@mjftw does that make sense to you? wheel isn't directly used, so I'd assume it should be installed when installing setuptools (or that's a bug in setuptools), what do you think?

PR #209, opened 2021-02-04
update package name for libusb-1.0-0 and update groupadd

mc18g13:
adding a user to a group would stop the script if the user was already in the group.
Also believe the package name for libusb added an extra -0


mjftw (approved):
Thanks very much for this. Looks good and all the CI is passing so good to go.

Please could you rebase to update your branch with master so I can merge?

PR #210, opened 2021-02-04
Fix: CommandRunner.run was ignorning timeout arg

mjftw:


Pixep (approved):
Fixed the typecheck part. I remember fixing this and... I don't know where the fix actually go, or if it got lost haha. Probably got lost.

PR #211, opened 2021-02-04
Refactor pluma to allow yaml string configs

mjftw:
Split this out from https://github.com/Witekio/pluma-automation/pull/201.
This work should be ready to merge, and I'd like to get it in before too many changes happen that make it stale while I work on the REST side of things.

This is purely a refactor to move the side effects (filesystem and environment reads) up to the highest level.
This paves the way to having a different user interface to the CLI running with little effort.

### DIFF
--- a/pluma/__main__.py
+++ b/pluma/__main__.py
@@ -102,8 +102,9 @@ def main():
         command = args.command
 
         if command in [RUN_COMMAND, CHECK_COMMAND, TESTS_COMMAND]:
-            pluma_context, tests_config = Pluma.create_context_from_files(tests_config_path,
-                                                                          target_config_path)
+            env_vars = dict(os.environ)
+            pluma_context, tests_config = Pluma.create_context_from_files(
+                tests_config_path, target_config_path, env_vars)
 
             if command == RUN_COMMAND:
                 success = Pluma.execute_run(pluma_context, tests_config)
--- a/pluma/__main__.py
+++ b/pluma/__main__.py
@@ -100,14 +100,20 @@ def main():
 
     try:
         command = args.command
-        if command == RUN_COMMAND:
-            success = Pluma.execute_run(tests_config_path, target_config_path)
-            exit(0 if success else 1)
-        elif command == CHECK_COMMAND:
-            Pluma.execute_run(tests_config_path, target_config_path,
-                              check_only=True)
-        elif command == TESTS_COMMAND:
-            Pluma.execute_tests(tests_config_path, target_config_path)
+
+        if command in [RUN_COMMAND, CHECK_COMMAND, TESTS_COMMAND]:
+            pluma_context, tests_config = Pluma.create_context_from_files(tests_config_path,
+                                                                          target_config_path)
+
+            if command == RUN_COMMAND:
+                success = Pluma.execute_run(pluma_context, tests_config)
+                exit(0 if success else 1)
+            elif command == CHECK_COMMAND:
+                Pluma.execute_run(pluma_context, tests_config,
+                                check_only=True)
+            elif command == TESTS_COMMAND:
+                Pluma.execute_tests(tests_config_path)
+
         elif command == CLEAN_COMMAND:
             Pluma.execute_clean(args.force)
         elif command == VERSION_COMMAND:
--- a/pluma/cli/api.py
+++ b/pluma/cli/api.py
@@ -29,14 +29,10 @@ class Pluma:
                 CTestsProvider(), DeviceActionProvider()]
 
     @staticmethod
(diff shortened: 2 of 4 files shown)


pluma/cli/api.py
```diff
@@ -81,30 +79,64 @@ def execute_clean(force: bool = False):
...
+    def create_context_from_files(tests_config_path: str, target_config_path: str,
+                                  substitute_vars: Optional[Dict[str, Any]] = None
+                                  ) -> Tuple[PlumaContext, TestsConfig]:
+        substitute_vars = substitute_vars or {}
+        target_config_opts = Pluma.load_config_file(target_config_path, 'Target config', substitute_vars)
+        context = Pluma.create_target_context(target_config_opts, substitute_vars)
+        tests_config_opts = Pluma.load_config_file(tests_config_path, 'Tests config', context.variables)
+        tests_config = Pluma.create_tests_config(tests_config_opts, context)
+
+        return context, tests_config
```
Pixep, line 91: Only returning the tests_config here, but not the target_config? Is the context enough?

pluma/cli/api.py
```diff
@@ -81,30 +79,64 @@ def execute_clean(force: bool = False):
...
+        substitute_vars = substitute_vars or {}
+        target_config_opts = Pluma.load_config_yaml(target_config_yaml, 'Target config', substitute_vars)
+        context = Pluma.create_target_context(target_config_opts, substitute_vars)
+        tests_config_opts = Pluma.load_config_yaml(tests_config_yaml, 'Tests config', context.variables)
+        tests_config = Pluma.create_tests_config(tests_config_opts, context)
+
+        return context, tests_config
+
+    @staticmethod
+    def load_config_yaml(yaml_str: str, name: str, extra_vars: Optional[Dict[str, Any]] = {}
```
Pixep, line 114: Probably needs a docstring or something to differentiate it from `load_config_file`. At first glance, I would expect this to do the same.

pluma/cli/config.py
```diff
@@ -184,28 +184,40 @@ def preprocess(self, raw_config: str) -> str:
...
-                           preprocessor: ConfigPreprocessor = None) -> Configuration:
-        return Configuration(PlumaConfig.load_yaml(name, config_path, preprocessor))
+    def load_configuration_file(name: str, config_path: str,
+                           preprocessor: Optional[ConfigPreprocessor] = None) -> Configuration:
+        return Configuration(PlumaConfig.load_yaml_file(name, config_path, preprocessor))
 
     @staticmethod
-    def load_yaml(name: str, yaml_file_path: str,
-                  preprocessor: ConfigPreprocessor = None) -> dict:
+    def load_configuration_yaml(name: str, config_yaml: str,
```
Pixep, line 192: Same regarding the need for a docstring, or a way to differentiate load_configuration_file and load_configuration_yaml from the outside. Possibly the name

pluma/cli/config.py
```diff
@@ -184,28 +184,40 @@ def preprocess(self, raw_config: str) -> str:
...
-        except ParserError as e:
+        except Exception as e:
             raise ConfigurationError(
-                f'Error while parsing {name} "{yaml_file_path}":'
+                f'An error occurred while opening {name} "{yaml_file_path}":'
                 f'{os.linesep}{e}') from e
-        except Exception as e:
+
+    @staticmethod
+    def load_yaml_str(name: str, yaml_str: str,
```
Pixep, line 212: Yep, good idea!

Pixep (changes requested):
Nice, thanks for the effort, and additional tests :). A few remarks, one bug to fix I think.

pluma/cli/api.py
mjftw: The `target_config` is consumed while making the `context`, so should be empty at the return point.
In any case it's never used again so we should be okay to discard it.

pluma/cli/api.py
mjftw: Good point, I'll see about just making the names longer and adding docstrings

pluma/__main__.py
```diff
@@ -100,14 +100,21 @@ def main():
...
+                tests_config_path, target_config_path, env_vars)
+
+            if command == RUN_COMMAND:
+                success = Pluma.execute_run(pluma_context, tests_config)
+                exit(0 if success else 1)
+            elif command == CHECK_COMMAND:
+                Pluma.execute_run(pluma_context, tests_config,
+                                check_only=True)
+            elif command == TESTS_COMMAND:
+                Pluma.execute_tests(tests_config_path)
```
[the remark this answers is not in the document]
mjftw: Oh yes you're right. I've fixed it for now but we can look at what test is missing.

PR #214, opened 2021-02-05
install.sh: configure adb support if required

nelsmore:


install.sh
```diff
@@ -132,10 +139,19 @@ $SUDO apt update && \
 if [ "$answer_all" ==  "n" ]; then
     echo "Skipping optional config..."
 elif [ "$answer_all" ==  "y" ]; then
+    support_android
     support_serial
     support_sdwire
     support_uhubctl
 else
+    echo
+    read -p 'Android adb support required? (Y/n): ' require_android
```
mjftw, line 148: Would be better to put this as defaulting to No rather than defaulting to Yes since it's an uncommon use case AFAIK.

mjftw (changes requested):
Thanks for the add 👍🏼 

I've not actually used Pluma with Android before, what is the use case you've got?
I'm a little hesitant to add Android support to the install script as the Pluma code doesn't require Android at this point AFAIK, in as much as none of the Pluma functionalities rely on `android-tools-adb` etc.
Definitely useful to have for people that want Android support, but feels more like a manual install step for their specific setup to me.

@Pixep what's your take on this?

install.sh
nelsmore: Ok that's no problem.  Do you have any preference on when this question is asked?  I added it first to maintain alphabetic order, but if its less common than adding it last may be better?

nelsmore:
> Thanks for the add 👍🏼
> 
> I've not actually used Pluma with Android before, what is the use case you've got?
> I'm a little hesitant to add Android support to the install script as the Pluma code doesn't require Android at this point AFAIK, in as much as none of the Pluma functionalities rely on `android-tools-adb` etc.
> Definitely useful to have for people that want Android support, but feels more like a manual install step for their specific setup to me.
> 
> @Pixep what's your take on this?

I have a use-case that requires Android and have begun adding some basic adb/Android support in  #216

